PR #7, opened 2016-11-07
Make LottieAnimationView reusable

ngsilverman:
Also added `Cycle` to the sample app that cycles through all the available animation using the same `LottieAnimationView`.


Edward601 (changes requested):
I won't to string everything back to real data

PR #20, opened 2017-02-01
Miscellaneous gradle and project fixes

felipecsl:
* Added travis config file
* Updated a couple dependencies
* Added missing gradle wrapper jar
* Removed unnecessary apt plugin

@gpeal

PR #21, opened 2017-02-01
Reduce down to just the needed support lib

Jawnnypoo:
Currently, Lottie is relying on app-compat, when in reality, it only needs `support-annotations` and `support-core-ui`

Though most apps probably already rely on appcompat, this will reduce bloat for those that do not.


Jawnnypoo:
Looks like https://github.com/airbnb/lottie-android/pull/20 switched to AppCompatImageView, so disregard this

PR #48, opened 2017-02-02
Prevent parent layers from overwriting child layer's alpha

gpeal:


gpeal:
@elihart @felipecsl @felipecsl

PR #49, opened 2017-02-02
Invalidate LottieDrawable with itself so that it can be verified

### LINKED ISSUE #37: Trying to use LottieDrawable: animation doesn't start
Hi!
I just tried Lottie and I could make `LottieAnimationView` work with no problems.

However, I cannot make `LottieDrawable` work, and I wonder what I am doing wrong.

Here is what I do:
```
        LottieDrawable lottieDrawable = new LottieDrawable();
        mMainBinding.imgLottie.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
        mMainBinding.imgLottie.setImageDrawable(lottieDrawable);
        LottieComposition.fromAssetFileName(this, "TwitterHeart.json", (composition) -> {
            lottieDrawable.setComposition(composition);
            lottieDrawable.loop(true);
            lottieDrawable.playAnimation();
        });
```

The result is I only see the first frame - the animation doesn't start.
I looked at the source of `LottieAnimationView` to compare, and didn't see any major difference.
Any idea?

gpeal:
Previously, LottieDrawable would invalidate itself with whatever layer changed. This would fail verification in [View](https://github.com/android/platform_frameworks_base/blob/master/core/java/android/view/View.java#L17951).

This bubbles up invalidate calls so that the drawable will invalidate itself with `LottieDrawable`.

Fixes #37

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/layers/AnimatableLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/layers/AnimatableLayer.java
@@ -106,6 +106,13 @@ public class AnimatableLayer extends Drawable {
         canvas.restoreToCount(saveCount);
     }
 
+    @Override
+    public void invalidateSelf() {
+        if (parentLayer != null) {
+            parentLayer.invalidateSelf();
+        }
+    }
+
     int saveCanvas(@Nullable Canvas canvas) {
         if (canvas == null) {
             return 0;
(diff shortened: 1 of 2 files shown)


gpeal:
@felipecsl @elihart @ngsilverman

elihart:
lgtm

PR #54, opened 2017-02-03
Flatten all classes into single package

felipecsl:
Moved all classes into a single flat package structure. This will make it easier to avoid accidentally leaking classes to users and to start breaking it down into smaller/self contained modules.
Next step is to decouple the parsing logic into a new module.
A lot of these changes are actually formatting fixes (wrapping due to right margin limit reached)

Part of #3


LottieSample/src/main/java/com/airbnb/lottie/samples/AnimationFragment.java
```diff
@@ -71,8 +71,8 @@ static AnimationFragment newInstance() {
   @BindView(R.id.animation_name) TextView animationNameView;
 
   @Nullable
-  @Override
-  public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+  @Override public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
```
gpeal: @felipecsl What's the convention for the method annotations here?

PR #95, opened 2017-02-13
Update README.md

dakaugu:


gpeal:
@dakaugu using non-explicit versions in gradle is not recommend. I'm about to push 1.0.3 so I'll update the readme to reflect that.

dakaugu:
@gpeal got it. When i get the warning in gradle, it's sort of a reminder for me to check the latest release and add the correct version number. That was the intention. The reason I suggested this change is because I faced issue#36 without knowing there was an update.
Thanks for the awesome library. :)

gpeal:
@dakaugu glad you like it :)
I just pushed 1.0.3 so it should be live in the next few hours whenever it propagates maven's caches.

PR #128, opened 2017-02-19
Converted all json to use opt* instead of get*

### LINKED ISSUE #123: Lot of unnecessary try/catch blocks in parsing code
Issue: So, exceptions are expensive — https://shipilev.net/blog/2014/exceptional-performance/ . In parsing code there are lot of `try/catch` blocks handling `JSONException` with empty `catch` blocks. In fact, these blocks are for reading optional fields. We can use `opt*` methods of `JSONObject` instead of `get*`, behavior under the hood is the same, but `get*` tries to raise an exception, while `opt*` falls back to null/default value/provided value.

Impact: slower parsing (ok, it won't be as fast as it could be in plugged Jackson/Gson/you name it), bigger memory consumption during parsing, larger codebase.

Proposed solution: replace `get*` with `opt*` where possible — ideally everywhere.

gpeal:
* The parsing code is now cleaner.
* There is a slight performance improvement using opt* instead of get*. I did some tests and didn't see a significantly measurable improvement but for large files or on certain devices, it may make a difference.

Fixes #123 
#3

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/AnimatableColorValue.java
+++ b/lottie/src/main/java/com/airbnb/lottie/AnimatableColorValue.java
@@ -3,20 +3,19 @@ package com.airbnb.lottie;
 import android.graphics.Color;
 
 import org.json.JSONArray;
-import org.json.JSONException;
 import org.json.JSONObject;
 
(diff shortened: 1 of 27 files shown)


gpeal:
@0legg fyi

PR #132, opened 2017-02-20
Added support for subtract masks

gpeal:


alexpersian:
Will this support be added to iOS as well?

gpeal:
@alexpersian Yes but on a slightly delayed timeline. You can track the status over there.

PR #145, opened 2017-02-22
Decouple more json parsing logic

felipecsl:
This was a bit more involved as the JSON parsing in  `Keyframe` and `AnimatableValue` classes is a lot more involved.
All tests still pass but this change makes me a bit more nervous, @gpeal 
Please take a closer look if you can.

Part of #3


lottie/src/main/java/com/airbnb/lottie/Keyframe.java
```diff
@@ -52,62 +38,22 @@ static void setEndFrames(List<? extends Keyframe<?>> keyframes) {
...
 
   private final LottieComposition composition;
-  T startValue;
-  T endValue;
-  Interpolator interpolator;
-  @SuppressWarnings("WeakerAccess") float startFrame;
-  @SuppressWarnings("WeakerAccess") @Nullable Float endFrame = null;
-
-  public Keyframe(LottieComposition composition, float startFrame, float endFrame) {
+  @Nullable T startValue;
```
gpeal: Can these be final?

lottie/src/main/java/com/airbnb/lottie/Layer.java
```diff
@@ -244,30 +244,24 @@ static Layer newInstance(JSONObject json, LottieComposition composition) {
         preCompHeight = (int) (json.optInt("h") * composition.getScale());
       }
 
-      long inFrame = json.optLong("ip");
-      long outFrame = json.optLong("op");
+      float inFrame = json.optLong("ip");
```
gpeal: Hmm, I don't think these can be float values. It's getting a long from the json anyways.
Can you change the json to `optInt` and all of the vars to int and Integer?

lottie/src/main/java/com/airbnb/lottie/ColorFactory.java
```diff
@@ -0,0 +1,28 @@
+package com.airbnb.lottie;
+
+import android.graphics.Color;
+
+import org.json.JSONArray;
+
+class ColorFactory {
```
felipecsl: sure

lottie/src/main/java/com/airbnb/lottie/Keyframe.java
felipecsl: They are modified from `AnimatableIntegerValue`

lottie/src/main/java/com/airbnb/lottie/Layer.java
felipecsl: This means that `inOutKeyframes` needs to change to `List<Keyframe<Integer>>`. Is that correct?

gpeal: @felipecsl yeah that's correct.

Issue #149, opened 2017-02-22
Add a `==` check in setComposition


gpeal:

gpeal:
@elihart

elihart:
I'll start work on this.

PR #150, opened 2017-02-23
Only update lottie drawable if composition changes

elihart:
@gpeal 

Addresses https://github.com/airbnb/lottie-android/issues/149

Any ideas on the best way to test this? I looked at the test robot and it only sets a composition once, so that wouldn't catch any issues with a second composition not working correctly. I couldn't think of good way to test this without creating a new test approach. Seems like a robot method that sets the same composition twice, and then different compositions, on the same view would work.


gpeal:
@elihart yeah you can definitely just make a one off robot method that just tests this

PR #164, opened 2017-02-25
Add tests for changing compositions

elihart:
Adds a test for running one composition, and then setting a new composition on the same view and running that one.

Adds a second test for setting the same composition twice in a row on the same view and making sure it runs as expected the second time.

I refactored the robot a bit to reuse the test setup code.

@gpeal


LottieSample/src/androidTest/java/com/airbnb/lottie/TestRobot.java
```diff
@@ -26,11 +26,49 @@ static void testAnimation(MainActivity activity, String fileName, String imageAs
...
+    loadCompositionOnView(view, null, secondFile);
+    recordScreenshots(view, "testChangingCompositions_" + secondFile, DEFAULT_ANIMATED_PROGRESS);
+
+    cleanUpView(view, activity);
+  }
+
+  static void testSettingSameComposition(MainActivity activity, String fileName) {
+    final LottieAnimationView view = new LottieAnimationView(activity);
+    loadCompositionOnView(view, null, fileName);
+    recordScreenshots(view, "same_composition_first_run_" + fileName, DEFAULT_ANIMATED_PROGRESS);
```
gpeal: nit: casing is different between this and your other test

PR #207, opened 2017-03-20
Ability to set color filter

seanfchan:
To make the composition more dynamic in regards to color we give the user of the library to ability to color specific layers dynamically.
This is akin to setting a color filter on drawables.


seanfchan:
I ran the screen shot tests and it seemed to run ok.
I haven't added a test yet. Will add when I get the right composition.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -36,6 +39,7 @@
   private float scale = 1f;
   private float progress = 0f;
 
+  @NonNull Map<String, ColorFilter> layerNameColorFilterMap;
```
gpeal: nit: anything that's not explicitly `@Nullable` is assumed to be `@NonNull`

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -47,6 +51,7 @@
   private int alpha = 255;
 
   @SuppressWarnings("WeakerAccess") public LottieDrawable() {
+    layerNameColorFilterMap = new HashMap<>();
```
gpeal: You can init this inline and make it final.

lottie/src/main/java/com/airbnb/lottie/DrawingContent.java
```diff
@@ -1,10 +1,19 @@
...
 interface DrawingContent extends Content {
   void draw(Canvas canvas, Matrix parentMatrix, int alpha);
   void getBounds(RectF outBounds, Matrix parentMatrix);
+
+  /**
+   * To color a specific layer in the composition
+   * @param name name of layer to apply the color filter to
+   * @param colorFilter the color filter to apply
+   */
+  void setColorFilter(@Nullable String name, @Nullable ColorFilter colorFilter);
```
gpeal: Why is name `@Nullable`. Maybe null should have it apply to all layers?

seanfchan: haha i was thinking about this as well.
I thought it might be a bit confusing for users if it just applied to all layers.

can definitely do this

gpeal: @seanfchan Just make sure it's clear in the javadoc. Can you also copy the javadoc to `LottieDrawable` and `LottieAnimationView`?

seanfchan:
1. Check, `addColorFilter` is a much better description
2. Just wanted to go over what needs to be done for this:
    * Would need to add name parsing for each `shapes`.
    * Applying of the color filter should remain on the specified layer, which means we would only search for the content name on the specified layer
    * If layer is specified but content name is not specified we would apply the color filter to the whole layer
    * If layer is not specified all layers apply the color filter

lottie/src/main/java/com/airbnb/lottie/DrawingContent.java
seanfchan: will do

gpeal:
@seanfchan Sounds good! It might be a little clearer if you just overload bullets 2 and 3 into 2 different methods like:
`addColorFilterToLayer(@Nullable String layerName, @Nullable ColorFilter colorFilter)`
and:
`addColorFilterToContent(String layerName, String contentName, @Nullable ColorFilter colorFilter)`
What do you think?

seanfchan:
:+1: sounds good to me

gpeal:
@seanfchan Ping me when this is ready for review.

seanfchan:
will do, want to get this into v2

gpeal:
@seanfchan Yup, sounds good to me. It'll be a while before 2.0.0 final mainly just to keep things in sync with iOS which doesn't have many of the 2.0 features yet.

seanfchan:
@gpeal i tested it on a test composition and it works in the following 3 cases
- the whole composition
- on a specific layer
- on a specific composition content

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -194,6 +214,54 @@ private void clearComposition() {
     // Do nothing.
   }
 
+  /**
+   * Add a color filter to specific content on a specific layer.
+   * @param layerName non null, name of the layer where the supplied content name lives
+   * @param contentName non null, name of the specific content that the color filter is to be
+   *                    applied
+   * @param colorFilter the color filter
```
gpeal: Same javadocs as interface.
Alternatively, you can only have the javdoc here and use `@see LottieDrawable#add*` as the javadoc in the DrawingContent interface

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -409,4 +477,46 @@ private ImageAssetBitmapManager getImageAssetBitmapManager() {
...
+    @Nullable final String contentName;
+    @Nullable final ColorFilter colorFilter;
+    private final int hashCode;
+
+    ColorFilterData(String layerName, @Nullable String contentName,
+        @Nullable ColorFilter colorFilter) {
+      this.layerName = layerName;
+      this.contentName = contentName;
+      this.colorFilter = colorFilter;
+      hashCode = (this.layerName + this.contentName).hashCode();
```
gpeal: 3 questions here:
1) Do you need equals and hashCode for this?
2) Why are you computing this in the constructor?
3) If you do need this, you can just move it to the hashCode function but the proper way to do this would be:
```
int hashCode = 17;
hashCode = hashCode * 31 * layerName.hashCode();
hashCode = hashCode * 31 * contentName.hashCode();
```
Then you don't have to do a string allocation either.

lottie/src/main/java/com/airbnb/lottie/ShapeFill.java
```diff
@@ -8,11 +8,13 @@
 class ShapeFill {
   private final boolean fillEnabled;
   private final Path.FillType fillType;
+  @Nullable private final String name;
   @Nullable private final AnimatableColorValue color;
   @Nullable private final AnimatableIntegerValue opacity;
-
-  private ShapeFill(boolean fillEnabled, Path.FillType fillType,
+
```
gpeal: nit: extra space

lottie/src/main/java/com/airbnb/lottie/BaseLayer.java
```diff
@@ -325,7 +326,17 @@ private void buildParentLayerListIfNeeded() {
     }
   }
 
+  @Nullable @Override public String getName() {
```
seanfchan: it is possible when parsing the json, the return value can be null

It is just to be clear to future contributors that a null value is possible

lottie/src/main/java/com/airbnb/lottie/CircleShape.java
```diff
@@ -1,14 +1,18 @@
 package com.airbnb.lottie;
 
 import android.graphics.PointF;
+import android.support.annotation.Nullable;
 
 import org.json.JSONObject;
 
 class CircleShape {
+  @Nullable private final String name;
```
seanfchan: yeah it would return empty string.

ok will change

lottie/src/main/java/com/airbnb/lottie/BaseLayer.java
seanfchan: will change

lottie/src/main/java/com/airbnb/lottie/CompositionLayer.java
```diff
@@ -106,4 +108,16 @@ boolean hasMatte() {
     }
     return false;
   }
+
+  @Override
+  public void addColorFilter(@Nullable String layerName, @Nullable String contentName,
```
seanfchan: i am not exactly sure what you are asking for here.

create another function such as `public void addColorFilter(@Nullable ColorFilter colorFilter)`?

gpeal: @seanfchan You can nest compositions (called precomps). That will create a CompositionLayer inside of the original CompositionLayer. You should be able to specify the precomp layer as the layer to get tinted and then it should tint all layers within the precomp. Does that make sense?

lottie/src/main/java/com/airbnb/lottie/ContentGroup.java
```diff
@@ -96,6 +99,23 @@
     lottieDrawable.invalidateSelf();
   }
 
+  @Override
+  public void addColorFilter(@Nullable String layerName, @Nullable String contentName,
+      @Nullable ColorFilter colorFilter) {
+    for (int i = 0; i < contents.size(); ++i) {
+      final Content content = contents.get(i);
+      if (content instanceof DrawingContent) {
+        if (layerName == null || contentName == null || contentName.equals(content.getName())) {
```
seanfchan: i'm going to split up this logic to make it more clear

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -135,6 +136,35 @@ private void init(@Nullable AttributeSet attrs) {
     super.setImageDrawable(drawable);
   }
 
+  /**
+   * Add a color filter to specific content on a specific layer.
+   * @param layerName non null, name of the layer where the supplied content name lives
+   * @param contentName non null, name of the specific content that the color filter is to be
+   *                    applied
+   * @param colorFilter the color filter
```
seanfchan: yes, i will add a cleanAllColorFilter method as well

seanfchan: yes

lottie/src/main/java/com/airbnb/lottie/CompositionLayer.java
seanfchan: ok i did understand your original point, but I think the current code supports that already. As long as the layerName is the precomp name.

seanfchan:
@gpeal back to you

lottie/src/main/java/com/airbnb/lottie/ContentGroup.java
```diff
@@ -96,6 +99,31 @@
...
+  }
+
+  @Override public void addColorFilter(@Nullable String layerName, @Nullable String contentName,
+      @Nullable ColorFilter colorFilter) {
+    for (int i = 0; i < contents.size(); i++) {
+      final Content content = contents.get(i);
+      if (content instanceof DrawingContent) {
+        final DrawingContent drawingContent = (DrawingContent) content;
+
+        if (layerName == null) {
```
gpeal: You can think of contents as children of a layer if you're not that familiar with After Effects. Given that, once you are in a DrawingContent (that doesn't extend BaseLayer), layerName isn't relevant.
CompositionLayer is the only thing that should need layerName.

lottie/src/main/java/com/airbnb/lottie/DrawingContent.java
```diff
@@ -1,10 +1,21 @@
...
 import android.graphics.RectF;
+import android.support.annotation.Nullable;
 
 interface DrawingContent extends Content {
   void draw(Canvas canvas, Matrix parentMatrix, int alpha);
   void getBounds(RectF outBounds, Matrix parentMatrix);
+
+  /**
+   * Add a color filter to specific content on a specific layer.
+   * @param layerName name of the layer where the supplied content name lives
```
gpeal: or null for all layers

lottie/src/main/java/com/airbnb/lottie/DrawingContent.java
```diff
@@ -1,10 +1,21 @@
...
+   * @param contentName name of the specific content that the color filter is to be applied
```
gpeal: or null to include all content that matches the layer param

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -194,6 +213,53 @@ private void clearComposition() {
...
+  /**
+   * Private method to capture all color filter additions.
+   * There are 3 different behaviors here.
+   * 1. layerName is null. All layers supporting color filters will apply the passed in color filter
+   * 2. layerName is not null, contentName is null. This will apply the passed in color filter
+   *    to the whole layer
+   * 3. layerName is not null, contentName is not null. This will apply the pass in color filter
+   *    to a specific composition content.
+   */
+  private void addColorFilterInternal(@Nullable String layerName, @Nullable String contentName,
```
gpeal: It may be a good idea to add a clearColorFilters method in case they apply a totally unrelated animation to the same view

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -409,4 +475,47 @@ private ImageAssetBitmapManager getImageAssetBitmapManager() {
...
+    @Nullable final ColorFilter colorFilter;
+
+    ColorFilterData(String layerName, @Nullable String contentName,
+        @Nullable ColorFilter colorFilter) {
+      this.layerName = layerName;
+      this.contentName = contentName;
+      this.colorFilter = colorFilter;
+    }
+
+    @Override public int hashCode() {
```
gpeal: Do you need hashCode and equals?

seanfchan: Since these objects are being put into a set, I believe I do need these. Allows LottieDrawble to not perform unnecessary work.

seanfchan:
@gpeal back to you

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
gpeal: Ah right, they're in a set. Forgot about that.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -168,6 +177,16 @@ private void buildCompositionLayer() {
         this, Layer.Factory.newInstance(composition), composition.getLayers(), composition);
   }
 
+  private void applyColorFilters() {
+    if (compositionLayer == null) {
+      return;
+    }
+
+    for (ColorFilterData data : colorFilterData) {
```
gpeal: Just in case somebody calls setColorFilter in an update listener, can you make this a fori loop so it doesn't create an iterator object for each call?

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -194,6 +213,61 @@ private void clearComposition() {
...
+   * There are 3 different behaviors here.
+   * 1. layerName is null. All layers supporting color filters will apply the passed in color filter
+   * 2. layerName is not null, contentName is null. This will apply the passed in color filter
+   *    to the whole layer
+   * 3. layerName is not null, contentName is not null. This will apply the pass in color filter
+   *    to a specific composition content.
+   */
+  private void addColorFilterInternal(@Nullable String layerName, @Nullable String contentName,
+      @Nullable ColorFilter colorFilter) {
+    colorFilterData.add(new ColorFilterData(layerName, contentName, colorFilter));
```
gpeal: If you call setColorFilterw with a ColorFilter then try to clear it with null, the old data will still be there. This doesn't really clear it.

lottie/src/main/java/com/airbnb/lottie/Content.java
```diff
@@ -3,5 +3,7 @@
 import java.util.List;
 
 interface Content {
+  String getName();
```
gpeal: Is there a specific reason this needs to be on Content rather than DrawingContent other than consistency?

seanfchan: my understanding is that all content can have a name. just put it here to be consitent

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
seanfchan: well it would make that entry have a null colorFilter because colorFilterData is a Set. But making the removal explicit will make it so we don't do extra work

lottie/src/main/java/com/airbnb/lottie/Content.java
gpeal: It can, just wanted to make sure I understood your intention there since it doesn't seem like any PathContent names are used

gpeal:
@seanfchan Looks good! I'll +1 and merge it after the fori loop change

seanfchan:
@gpeal addressed the latest comments, and rebase off master

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -168,6 +177,18 @@ private void buildCompositionLayer() {
         this, Layer.Factory.newInstance(composition), composition.getLayers(), composition);
   }
 
+  private void applyColorFilters() {
+    if (compositionLayer == null) {
+      return;
+    }
+
+    ColorFilterData[] data = (ColorFilterData[]) colorFilterData.toArray();
```
gpeal: @seanfchan Why do you need to do toArray vs just iterating instead of
`for (int i = 0; i < colorFilterData.size(); i++)`?
toArray actually allocates a new array (https://android.googlesource.com/platform/libcore/+/refs/heads/master/ojluni/src/main/java/java/util/Arrays.java#3138)

seanfchan: @gpeal colorFilterData is a Set. As far as I am aware there are only 2 ways to iterate the contents, iterator or create the array. I can change it to a List if you feel that is more appropriate. I don't think the list of color filters would get that huge anyway

seanfchan: i should clarify that the code you presented is definitely possible, but there is not a method to retrieve the context with just the index without the array

gpeal:
Let's do the iterator then. Sorry for the churn!

seanfchan:
np

seanfchan:
@gpeal done

gpeal:
@seanfchan Just pushed beta3 to maven

PR #212, opened 2017-03-21
chore: update README to indicate playAnimation

kaushikgopal:
(suggestion for us lazy copy pasters) if animation file specified through xml it starts playing automatically, thus the true equivalent to setting this in java, would require `playAnimation`


kaushikgopal:
sorry @gpeal - couldn't assign you to the PR after the fact 😬

gpeal:
Hah, thanks!

PR #219, opened 2017-03-24
add color filter documentation to README

seanfchan:


README.md
```diff
@@ -121,6 +121,21 @@ animator.start();
 animationView.cancelAnimation();
 ```
 
+You can add a color filter to the whole animation, a specific layer, or specific content within a layer:
```
gpeal: Can you add a sentence that indicates that the content includes things like fill, stroke, or group, not a shape?

README.md
```diff
@@ -121,6 +121,21 @@ animator.start();
...
+```java
+final ColorFilter colorFilter = new PorterDuffColorFilter(Color.RED, PorterDuff.Mode.SRC_ATOP);
```
gpeal: Do you think we should create something like `SimpleColorFilter` that extends PorterDuffColorFilter and sets the mode. I wonder if many people will get confused by the xfer mode.

seanfchan: I will make one, but I feel that if an android dev is dealing with drawables and color filters they should learn which xfer mode suits their needs. 

I can see how this could generate the interest the learn other modes though.

seanfchan:
@gpeal back to you

README.md
```diff
@@ -121,6 +121,24 @@ animator.start();
...
+// Adding a color filter to a specific layer
+animationView.addColorFilterToLayer("hello_layer", colorFilter);
+
+// Adding a color filter to specfic content on the "hello_layer"
+animationView.addColorFilterToContent("hello_layer", "hello", colorFilter);
+
+// Clear all color filters
+animationView.clearColorFilters();
+```
+Note: Color filters are only available for layers such as Image layer and Solid layer as well as content that includes fill, stroke, or group content. Shapes like Polystar or Rectangles are currently not supported.
```
gpeal: @seanfchan. This isn't accurate. The way AE and Lottie work is that they walk up the tree of content from bottom to top. When it gets to a DrawingContent (like fill, stroke, etc), it then continues up and collects all of the PathContents above it (like Shape, Polystar, Rectangle, Ellipse, etc) and draws them all. 

This isn't immediately obvious to somebody not familiar with after effects and one's instinct may be to provide the name of the path content they want to fill. However, that won't and shouldn't work (partly because you can have an arbitrary number of fills and strokes on the same path) and you should specify the name of the drawing content instead.

It's not super easy to grasp/explain so I just want to make sure the documentation helps devs/designers understand that before being frustrated that it doesn't work.

seanfchan:
@gpeal back to you

gpeal:
@seanfchan I'm going to be on Fragmented tomorrow. If it comes up, can I mention that you guys have started playing around with Lottie?

seanfchan:
I love Fragmented. Sure you can mention us.

Best,
Sean

PR #231, opened 2017-03-28
Adding support for loading compositions with pre-scaling. 

BaloghTamas:
Adding support for loading compositions with pre-scaling. The composition gets loaded in a scaled way.  

In case of some compositions if you use fixed layout_width and layout_height attributes, elements get missing.  Try Lottiefiles.com - Postcard with fixed layout size (with a smaller one than the original composition).

In some cases this make a significant performance increase (if the composition is loaded into a much smaller view, it gets much faster)

Sample app menu extended with a fixed element boundary (100x100dp) to demonstrate.

Mixed sizes (Height:Fixed+Width:Wrap or Height:Wrap+Width:Fixed) not supported.


bmc08gt:
@gpeal adding reviewer

gpeal:
@BaloghTamas I think https://github.com/airbnb/lottie-android/pull/241 will solve your issue instead of this approach. It also doesn't break caching and can be set dynamically.

gpeal:
Closing per above comment

BaloghTamas:
I am afraid that #241 is not solving that issue, just expose the scaling functions.  
These scaling functions are based on scaling matrices, but handles the whole composition with its original scale, just renders them scaled. This can lead to performance issues and sometimes some parts of the compositions gets clipped. Please try with _Lottiefiles.com - Postcard_, with a fixed, quite small size (like ~75dp).

gpeal:
@BaloghTamas I just ran some tests prescaling the composition vs using the setScale API and the performance seems to be the same. Can you provide some evidence that this is not the case?

PR #234, opened 2017-03-29
fix crash if either layer name or content name were null

seanfchan:


seanfchan:
Say we only want to apply a color filter to a layer, the help method to apply a color filter on a layer will pass in null for the content. Either have these checks or we don't pass null down into apply color filter, but I think the later is more ambiguous.

PR #243, opened 2017-04-06
Use PorterDuffColorFilter in example in README

alexbirkett:
The example now illustrates how the color can be changed without having to implement a ColorFilter (which is non trivial as native code is involved)


gpeal:
Can you keep the original and add yours as a second example? It's helpful to know that you can use either the simple or or any arbitrary color filter.

PR #256, opened 2017-04-18
Add styleable attribute for whole animation color filter

ramonaharrison:
This PR adds a styleable attribute `lottie_colorFilter` to `LottieAnimationView` in order to simplify setup for assets that require different tints throughout the app.


digitalbuddha:
Could you add tests Pls.

gpeal:
This is great. If you want to add a test, you can inflate an xml file and then record a snapshot similar to what `TestRobot` does. Running `./gradlew recordMode screenshotTests` should add a png to your repo. Just include that and it will become the expected value for future tests.

Can you also update the readme?

gpeal:
@ramonaharrison Fyi ^

I'll also cut a new beta once this is merged

ramonaharrison:
Thanks! Added the README update.

I was running into some trouble inflating XML in androidTest to include `R` values from the Lottie library, but I think we have a working solution now. I'll finish up the tests this morning and push them up.

PR #157, opened 2017-02-23
Cancellable is returned by public functions, but isn't accessible

aaboyd:
You can do this ...
```
LottieComposition.Factory.fromJson(getResources(),
       new JSONObject(), new OnCompositionLoadedListener() {
           @Override
            public void onCompositionLoaded(LottieComposition composition) {
                ...
            }
       });
```

But you cannot do this ...
```
Cancellable compositionCancellable = LottieComposition.Factory.fromJson(getResources(),
       new JSONObject(), new OnCompositionLoadedListener() {
           @Override
            public void onCompositionLoaded(LottieComposition composition) {
                ...
            }
       });

compositionCancellable.cancel();
```

Also on the README.md the examples show the old style ```LottieComposition.fromJson``` as opposed to ```LottieComposition.Factory.fromJson```

PR #274, opened 2017-05-02
Make Cancellable interface visible

dustin-graham:
This is a regression since https://github.com/airbnb/lottie-android/pull/157
Cancellable is returned  from the composition factory but is inaccessible.


dustin-graham:
@gpeal I'm sorry, I don't know how to assign you as a reviewer as per the instructions.

PR #293, opened 2017-05-11
fix NullPointerException

imliujun:
If a null is returned in `ImageAssetDelegate.fetchBitmap()`,` recycleBitmaps()`  method Throws a NullPointerException

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/ImageAssetBitmapManager.java
+++ b/lottie/src/main/java/com/airbnb/lottie/ImageAssetBitmapManager.java
@@ -88,7 +88,10 @@ class ImageAssetBitmapManager {
     Iterator<Map.Entry<String, Bitmap>> it = bitmaps.entrySet().iterator();
     while (it.hasNext()) {
       Map.Entry<String, Bitmap> entry = it.next();
-      entry.getValue().recycle();
+      Bitmap value = entry.getValue();
+      if (value != null) {
+        value.recycle();
+      }
       it.remove();
     }
   }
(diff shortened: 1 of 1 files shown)


gpeal:
@imliujun Why is it null in the first place? Are you ever passing it a null bitmap to begin with?

gpeal:
@imliujun I think the correct fix is to not put the bitmap in the map if the delegate returns null in `bitmapForId`

imliujun:
If the image is downloaded from the network, it is possible to download or decompress the failure, causing decode bitmap to be null

gpeal:
@imliujun Makes sense. Can you guard against putting a null value in the map in the first place rather then when it is retrieved?

imliujun:
```java
public Bitmap fetchBitmap(LottieImageAsset lottieImageAsset) {
        Bitmap bitmap = null;
        BitmapFactory.Options opts = new BitmapFactory.Options();
        opts.inScaled = true;
        opts.inDensity = 160;
        String path = mImageDir + lottieImageAsset.getFileName();

        if (WebpSupportStatus.sIsWebpSupportRequired) {
            if (mWebpBitmapFactory == null) {
                mWebpBitmapFactory = new WebpBitmapFactoryImpl();
            }
            bitmap = mWebpBitmapFactory.decodeFile(path, opts);
        }
        if (bitmap == null) {
            bitmap = BitmapFactory.decodeFile(path, opts);
            if (bitmap == null) {
                bitmap = Bitmap.createBitmap(1, 1, Bitmap.Config.ARGB_4444);
            }
        }
        return bitmap;
    }
```

I am now going to avoid a null pointer by returning a 1 pixel bitmap, but will Lottie do some checks better?

gpeal:
@imliujun no, if the delegate return null in bitmapForId then just don't put it in the map. bitmapForId can return null.

imliujun:
@gpeal Then please close this pull request, please use the correct way to solve this problem bars

gpeal:
@imliujun Would you like to fix this PR or would you like me to do it separately?

imliujun:
I will submit a request for you to check, thanks

PR #307, opened 2017-05-23
Set scale.

howard0020:
Set scale through xml.
Keep previously set scale.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -118,6 +118,10 @@ public class LottieAnimationView extends AppCompatImageView {
       addColorFilter(new SimpleColorFilter(ta.getColor(
           R.styleable.LottieAnimationView_lottie_colorFilter, Color.TRANSPARENT)));
     }
+    if(ta.hasValue(R.styleable.LottieAnimationView_lottie_scale)) {
+        lottieDrawable.setScale(ta.getFloat(R.styleable.LottieAnimationView_lottie_scale, 1f));
+    }
+
     ta.recycle();
 
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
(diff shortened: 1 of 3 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -118,6 +119,8 @@ private void init(@Nullable AttributeSet attrs) {
       addColorFilter(new SimpleColorFilter(ta.getColor(
           R.styleable.LottieAnimationView_lottie_colorFilter, Color.TRANSPARENT)));
     }
+
+    scale = ta.getFloat(R.styleable.LottieAnimationView_lottie_scale, Float.MAX_VALUE);
```
gpeal: Move the scale code around line 373 to a helper method and make that the default value here. Then you can remove it from below and just call `setScale(scale)`

howard0020: `composition` might be not be loaded, so it can't calculate the right scale with screen size.

howard0020:
@gpeal done

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -154,7 +154,9 @@ boolean enableMergePathsForKitKatAndAbove() {
     clearComposition();
     this.composition = composition;
     setSpeed(speed);
-    setScale(1f);
+    if(scale == null) {
```
gpeal: nit: space after if

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
gpeal: How about you just call setScale on `lottieDrawable` here and if the attribute doesn't exist, then set it to the logic that's currently below and removing it there. Since you solved the issue of the drawable overwriting the scale on setComposition, you can set it here and be done.

howard0020: Please bear with me here, are you suggesting something like this.

```
if(ta.hasValue(R.styleable.LottieAnimationView_lottie_scale)) {
  lottieDrawable.setScale(ta.getFloat(R.styleable.LottieAnimationView_lottie_scale, 1f));
}
```

When done loading `Composition` still try to set the drawable scale according to the screen size.
```
if (compWidth > screenWidth ||
        compHeight > screenHeight) {
      float xScale = screenWidth / (float) compWidth;
      float yScale = screenHeight / (float) compHeight;
      float maxScaleForScreen = Math.min(xScale, yScale);

      setScale(Math.min(maxScaleForScreen, lottieDrawable.getScale()));
...
}
```

gpeal: @howard0020 yeah let's do that

howard-lin-zocdoc:
@gpeal can you review again! Thanks.

howard0020:
@gpeal can you review again. Thanks!

howard0020:
@gpeal omg just realized what you meant. Done.

PR #346, opened 2017-06-22
Clear curves list of temporary shape before reuse.

kojiashida:
This is a fix for the crash problem we observed with our animation data (reproduces on the sample Android player too):

06-20 15:58:07.822 29935 29935 E AndroidRuntime: java.lang.IllegalStateException: Curves must have the same number of control points. This: 12» Shape 1: 13»    Shape 2: 13
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.ShapeData.interpolateBetween(ShapeData.java:58)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.ShapeKeyframeAnimation.getValue(ShapeKeyframeAnimation.java:19)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.ShapeKeyframeAnimation.getValue(ShapeKeyframeAnimation.java:7)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.BaseKeyframeAnimation.getValue(BaseKeyframeAnimation.java:109)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.ShapeContent.getPath(ShapeContent.java:54)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.FillContent.draw(FillContent.java:71)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.ContentGroup.draw(ContentGroup.java:185)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.ContentGroup.draw(ContentGroup.java:185)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.ShapeLayer.drawLayer(ShapeLayer.java:24)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.BaseLayer.draw(BaseLayer.java:162)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.CompositionLayer.drawLayer(CompositionLayer.java:75)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.BaseLayer.draw(BaseLayer.java:162)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at com.airbnb.lottie.LottieDrawable.draw(LottieDrawable.java:292)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at android.widget.ImageView.onDraw(ImageView.java:1316)
06-20 15:58:07.822 29935 29935 E AndroidRuntime: »      at android.view.View.draw(View.java:17185)

Thank you.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/ShapeKeyframeAnimation.java
+++ b/lottie/src/main/java/com/airbnb/lottie/ShapeKeyframeAnimation.java
@@ -13,6 +13,7 @@ class ShapeKeyframeAnimation extends BaseKeyframeAnimation<ShapeData, Path> {
   }
 
   @Override public Path getValue(Keyframe<ShapeData> keyframe, float keyframeProgress) {
+    tempShapeData.getCurves().clear();
     ShapeData startShapeData = keyframe.startValue;
     ShapeData endShapeData = keyframe.endValue;
 


gpeal:
@kojiashida This will cause a huge amount of object churn as new CubicCurveData objects will get created for every control point on every frame by hitting [this](https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/ShapeData.java#L60) code path. That path will also just blindly use the number of control point in shape 1 leaving you in an indeterminate state since the two shapes don't have the same number of control points.

The reason that this crashes is that I don't currently have a good solution for interpolating between two shapes that have a different number of control points so Lottie intentionally crashes.

My recommendation is:
1) Rework your animation so that shapes have a constant number of control points or
2) Help out with a solution for animating between two shapes that have a different number of control points.

kojiashida:
Thank you, Gabriel. I see that the simple minded patch will cause perf degradation.

For the fix, would you take a solution that avoids crash but may not reproduce how the animation looks in AfterEffects? I'm thinking of trying to just duplicate the last control point (in either the before or after shape) to match the number.

On the other hand, in our case shapeData1 and shapeData2 has the same number of vertices, but not the "curves" member variable. We might be able to fix our case just by adding necessary number of control points to the list in the code location you pointed out. Does that sound right to you?

gpeal:
@kojiashida Can you attach an animation that repros this? I wouldn't expect your shape data to be the same when curves is different. A single shape animation animates between shape data keyframes, all of which must have the same number of control points so it should never change.

Regarding duplicating the control point, my goal with Lottie is to be as faithful to AE as possible. The correct thing would be to have your designer actually update the animation. You can imagine a scenario in which the control point does actually matter and the resulting animation looks totally broken. This could be particularly bad if we were, for example, to alter a company logo inadvertently.

kojiashida:
@gpeal Understood on the importance of fidelity with AE. Given that, here's what I will do:

- I can't think of a simple fix so I'll just drop this PR and pursue art data side fix for our particular problem.

- The animation is for our product so I can't send you as is, but I'll ask our artist to make a trivial repro version that we can send it to you.

- I'll also get the AE file from our artist to see how this happened.

Thank you for your help.

Issue #352, opened 2017-06-27
Improve the keyframe value caching


gpeal:
https://github.com/airbnb/lottie-android/commit/d5f141dbbe6c7e1ab90e7ca94bb0b38af561fec5 doesn't work. Investigate.

azchohfi:
Yup, same wrong behavior on LottieUWP:
https://github.com/azchohfi/LottieUWP/commit/038222b6d63948aae2df3a3d4ea02df9e3c0129e#diff-3beb6887bf96b8d4981b98f00518ebcaR74

gpeal:
The regression is now fixed. I'll open future issues for perf improvements.

Issue #350, opened 2017-06-26
Add support for dynamic text


gpeal:
This enabled localization but requires drawing text from a font asset rather than glyphs

#349

Issue #173, opened 2017-03-02
Add support for text


KucherenkoIhor:
Hello. Can Lottie display text?? 
[single_buton.txt](https://github.com/airbnb/lottie-android/files/813581/single_buton.txt)
I have button with text "LET'S DO  IT" and text doesn't display.

0legg:
Can be done the same way image assets works utilizing Calligraphy (or just loading plain typefaces from assets)

tehfailsafe:
@0legg Can you elaborate? If I'm just using roboto, you're saying just drop the font file in the assets folder and use `lottieView.setImageAssetsFolder("assets");`?

gpeal:
Text glyph rendering now works! I'll open specific issues for further text enhancements and reference this one.

fentonbutcher:
Hi, I've tried exporting an animation via bodymovin that contains text and it still isn't displaying.  I've installed the release from a day ago. Any help would be appreciated

gpeal:
@fentonbutcher upgrade bodymovin to 4.9.0. You can also check composition.getWarnings(). It'll give you some hints.

fentonbutcher:
@gpeal where do I check the composition.getWarnings().?
Also, I have checked 'Gliphs  - checked converts fonts into shapes'. I still can't see any text. Have upgraded to the most recent 4.9.0 version

gpeal:
@fentonbutcher Also make sure you have the latest Lottie (2.1.0)

fentonbutcher:
We're using Lottie it for react native. Has that also been updated?

gpeal:
@fentonbutcher not yet but try explicitly putting 2.1.0 in your gradle file to see if you can force it to use a newer version.

fentonbutcher:
@gpeal is there a time frame for updating for react native?

fentonbutcher:
@gpeal just thought i'd say... thanks for the help so far :)

gpeal:
@fentonbutcher I pushed 1.2.0 for react native which supports text :)

fentonbutcher:
@gpeal is it possible to use dynamic text with this now as well in the same way as:

https://codepen.io/airnan/pen/ZLVJmq?editors=0010

gpeal:
@fentonbutcher Yes, use a text deleate

fentonbutcher:
@gpeal do I need to install a new version of bodymovin for after effects to get this to work?

fentonbutcher:
[...]

PR #353, opened 2017-06-27
Add support for rendering fonts and dynamic text

gpeal:
This support builds on text glyph support by supporting rendering custom fonts as well as custom text through a text delegate. This enables Lottie to be used with translations as well as interactive animations.

All text operations such as fills, strokes, tracking, etc are supported by this.

dynamic-text

#173
#349
#350
#352


Antiblanks:
Hey Gabriel we made a start on a iOS/Android project a few months ago where we looked at using Lottie but sidestepped it because of the lack of dynamic text support at the time. We're about to pickup that same project again and have noticed you guys have turned some attention to this more recently. Can you tell me where you are with this now both for iOS and Android and what your roadmap is for improving this feature? I need to decide whether to continue with the custom route we were going down or consider using Lottie again. If you are going to be supporting soon then what restrictions are there on using text, particularly effects with text; more basic stuff like animating drop shadows, glows etc to more complex stuff like morphing the shapes of letters. Let me know if you get the chance... Thanks in advance.

gpeal:
@Antiblanks I added basic text support on Android. You can find a list of supported features [here](http://airbnb.io/lottie/supported-features.html). You can also use a TextDelegate to do custom text.

iOS doesn't support text yet. You can file/search for an issue there for more info.

jalopez1986:
@gpeal How r u?, For when can you have support for this feature in ios?

gpeal:
@jalopez1986 It is on the roadmap.

Issue #431, opened 2017-08-11
SetScale not working


sinzen:
Hi,

I'm using lottie to do an animation. I try to resize the animation with : `animationView.setScale(0.9f);`

but isn't works.. here are my layout:
`<com.airbnb.lottie.LottieAnimationView
        android:id="@+id/animation_view"
        android:layout_width="100sp"
        android:layout_height="100sp"
        app:lottie_autoPlay="true"
        app:lottie_fileName="data.json"
        app:lottie_loop="true"
       />`

Thanks for your help

sinzen:
I'm back..
Is the property below is missed to scale the animation?
`scaleType=centerCrop` 
In my xml view I saw it here: [https://github.com/airbnb/lottie-android/issues/326](url)

gpeal:
@sinzen can you attach a sample project that repros this?

TheEnsis:
Me too.
I'm used file checked_done.json from lottiefile.com

sinzen:
Hi,
To solve the issue I tried another way and it works..
I took the aep file, and I réduction the border of composition into the limit of my animation..
Android reseize to get a an animation without pixels.
In my XML (Android proj) I fixe the size and that's it. I'm doesn't use setscale in the java class.
I can post the project if you need ?

gpeal:
@sinzen @TheEnsis v2.2.4 is now out

Issue #380, opened 2017-07-13
The Size is 2 times larger for wrap_content


talentqing:
LottieAnimationView use wrap_content for size, the actual size will be scaled 2 times.

for example:
PinJump.json size is 150x150
the actual is 2 times（screen size is 1080x1920）
![2](https://user-images.githubusercontent.com/30134668/28148475-8047a330-67b9-11e7-9a26-6cd501fab894.png)

the correct size
![image](https://user-images.githubusercontent.com/30134668/28148523-c1116324-67b9-11e7-9b50-a3514ade3e10.png)

float scale = res.getDisplayMetrics().density is no problem? this scale is used for dip, should the scale divide 2?

gpeal:
@talentqing Not sure I understand the question. Lottie converts all px values in AE to dps which will be ~2-3x depending on the device.

talentqing:
PinJump.json size is 150x150px
in picture 1, LottieAnimationView size is wrap_content in xml
in picture 2, LottieAnimationView size is 75dp in xml
picture 2 is correct size but picture 1 is not, so i think wrap_content has bug
in picture 2, 75dp correctly converts px value depending on the device

talentqing:
This is may be not a bug, main current ui making lottie in AE base 2x device.

gpeal:
@talentqing I'm not sure I understand the bug or the suggestion. Can you clarify?

maoruibin:
I meet same issue, my designer give me 120 * 120 json file,  and i attache it  to my LottieAnimationView, LottieAnimationView use wrap_content for size, my device is 3x, so LottieAnimationView's size is change 360 * 360,  it's no problem, animation is play normally,  but i want it 120 * 120, so i used setScale(1f/3f) method for LottieAnimationView, but animation is play error, Now, i want to know this issue is Json file error or not ? thanks @gpeal

GuillaumeLachaud:
I am not sure I can clarify what the original author of this task is meaning but I think I am seeing something very similar. It affects LottieAnimationView when using `wrap_content `. It got introduced somewhere between Lottie 2.0.0 and 2.2.0. I can also confirm https://github.com/airbnb/lottie-android/pull/461 fixes the issue.
As an alternative, using `.setScale(0.5f)` in our app solves the issue. But it used to be working properly on 2.0.0 without setting the scale explicitly.

See screens below taken from the sample app. 

**Lottie 2.2.3:**
[...]

PR #461, opened 2017-09-11
Lots of scaling improvements

gpeal:
Lottie should now respect scaling and scale types much better. Things like centerInside and centerCrop will now work as expected. The only limitation is that if you set the scale to larger than the canvas then it will only scale up to the maximum size for the current scale type (centerInside/centerCrop). If you change the canvas scaling from extraScale * extraScale then it will scale larger but it'll also end up scaling when you have a fixed width which you don't actually want. For now, I'm optimizing for the case where you want the animation to fit the view.

Fixes #461 #380 #431


jgautheron:
@gpeal Any chance you could update the `lottie-react-native` project to use `lottie-android` 2.2.5? Thanks!

gpeal:
@jgautheron yeah, I will

PR #499, opened 2017-10-12
Introduce a new lottie_rawRes XML attribute on LottieAnimationView

cyrilmottier:
The new API follows the “file from assets” API:
  - New `setAnimation(int)` and `setAnimation(int, CacheStrategy)` methods
  - If both `lottie_fileName` & `lottie_rawRes` are set, an `Exception` is thrown

Some notes behind this:
  - The main idea is to enable features like "Find usages" or "Unused resource lint" by introducing a hard link between the Java code and the Lottie animation file (via `R.java`).

Some points to be discussed:
  - I can't run the screenshot tests. I end up with a failure on `:LottieSample:pullScreenshots`
  - I followed the current logic for `View` state save/restoration etc. To be honest, I don't agree with the current implementation and consider `View` state should not include `animationName` and `animationResId` as these are generally reconstructed in the `Fragment`/`Activity` `View` inflation.


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -54,6 +55,10 @@
     Strong
   }
 
+  private static final SparseArray<LottieComposition> STRONG_RAW_REF_CACHE = new SparseArray<>();
+  private static final SparseArray<WeakReference<LottieComposition>> WEAK_RAW_REF_CACHE =
+      new SparseArray<>();
+
   private static final Map<String, LottieComposition> STRONG_REF_CACHE = new HashMap<>();
```
gpeal: Can you rename this to `STRONG_ASSET_REF_CACHE` and `WEAK_ASSET_REF_CACHE`?

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -100,9 +106,15 @@ private void init(@Nullable AttributeSet attrs) {
...
     defaultCacheStrategy = CacheStrategy.values()[cacheStrategy];
-    String fileName = ta.getString(R.styleable.LottieAnimationView_lottie_fileName);
-    if (!isInEditMode() && fileName != null) {
-      setAnimation(fileName);
+    if (!isInEditMode()) {
+      String fileName = ta.getString(R.styleable.LottieAnimationView_lottie_fileName);
+      if (fileName != null) {
+        setAnimation(fileName);
+      }
+      int rawResId = ta.getResourceId(R.styleable.LottieAnimationView_lottie_src, 0);
```
gpeal: Hmm, instead of immediately overwriting the asset if res is also set, how about something like:
```
String assetName = ...
@ResId int rawResId = ...
if (assetName != null && rawResId != 0) {
    throw new IllegalArgumentException("...");
} else if (assetName != null) {
    ...
} else if (rawResId != 0) {
    ...
}
```

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -220,6 +233,10 @@ public void clearColorFilters() {
     if (!TextUtils.isEmpty(animationName)) {
       setAnimation(animationName);
     }
+    animationResId = ss.animationResId;
+    if (animationResId > 0) {
```
gpeal: Check for `!= 0`

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -305,6 +322,55 @@ public void useHardwareAcceleration(boolean use) {
   }
 
   /**
+   * Sets the animation from a file in the raw directory.
+   * This will load and deserialize the file asynchronously.
+   * <p>
+   * Will not cache the composition once loaded.
+   */
+  public void setAnimation(int animationResId) {
```
gpeal: `@RawRes`

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -305,6 +322,55 @@ public void useHardwareAcceleration(boolean use) {
...
+
+  /**
+   * Sets the animation from a file in the raw directory.
+   * This will load and deserialize the file asynchronously.
+   * <p>
+   * You may also specify a cache strategy. Specifying {@link CacheStrategy#Strong} will hold a
+   * strong reference to the composition once it is loaded
+   * and deserialized. {@link CacheStrategy#Weak} will hold a weak reference to said composition.
+   */
+  public void setAnimation(final int animationResId, final CacheStrategy cacheStrategy) {
```
gpeal: `@RawRes`

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -324,6 +390,7 @@ public void setAnimation(String animationName) {
    */
   public void setAnimation(final String animationName, final CacheStrategy cacheStrategy) {
     this.animationName = animationName;
+    animationResId = -1;
```
gpeal: Set this to 0.

lottie/src/main/java/com/airbnb/lottie/LottieComposition.java
```diff
@@ -194,6 +194,14 @@ public static Cancellable fromAssetFileName(Context context, String fileName,
     }
 
     /**
+     * Loads a composition from a file stored in res/raw.
+     */
+    public static Cancellable fromRawFile(Context context, int resId,
```
gpeal: `@RawRes`

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
cyrilmottier: 👍

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
cyrilmottier: I looked at the rest of the class and it's not using support annotations. For instance the signature for `setImageResource` is `@Override public void setImageResource(int resId)`.

But I'll change that.

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
cyrilmottier: Let's stick to 0 for an invalid resource then :)

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
cyrilmottier: I agree.

cyrilmottier: But in your case `lottie_fileName` takes precedence. Is that something wanted?

Regarding the Exception, I would prefer sticking to only picking the preferred one (just like `android:padding` vs `android:paddingRight`). Could you please explain why you would go for an `Exception`?

I thought about an option that would simplify the use of the XML attribute but I didn't go into that path because it's really breaking the current API and the apps out there. The idea would be to get rid to `lottie_fileName` and replace it by `lottie_src` that could take either a `String` or a resource id to a raw file. Here is a quick list of the pros & cons I found:

Pros:
  - API similar to `ImageView`
  - Simpler API and no weird case with both set
 
Cons:
  - Break the current API

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -71,6 +77,7 @@
   private final LottieDrawable lottieDrawable = new LottieDrawable();
   private CacheStrategy defaultCacheStrategy;
   private String animationName;
+  private int animationResId;
```
gpeal: This should also be `@RawRes`

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
gpeal: @cyrilmottier Hmm, we could go with src and deprecate fileName and remove it in a future release.

I prefer failing loudly when you do unexpected behavior like set two conflicting APIs.

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
gpeal: @cyrilmottier It may be missing in a few cases but the intention is for them to be there. Feel free to add any you find that are missing!

cyrilmottier:
I've fixed the code according to the review.
I also added a WIP commit regarding the deprecation of `lottie_fileName` in favor of `lottie_src`. Unfortunately it looks like it's not such a great idea as problems occurred while implementing this. Here is the commit message that explains the main issues:

> Instead of having two different XML attributes depending to set the source file to LottieAnimationView, we only expose one: lottie_src. lottie_src accepts both references and direct strings. This has several drawbacks:
>
>  - It looks like it break Android Studio autocompletion (at least on 3.0 beta 7)
>  - There is no way to differentiate between a reference @raw/ and @string/. As a consequence, the current implementation currently considers @string/ as a reference and uses it to inflate the animation file which never ends well… I haven’t found a clean and simple way to workaround this yet.

I would be happy to have your feedback on this.

gpeal:
@cyrilmottier I think the approach in your commit is reasonable. Another approach is to keep `lottie_fileName` and add `lottie_src` or maybe call it `lottie_rawRes` for explicitness and to prevent us from having to deprecate or overload the attribute.

cyrilmottier:
Updated

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -54,8 +56,12 @@
     Strong
   }
 
-  private static final Map<String, LottieComposition> STRONG_REF_CACHE = new HashMap<>();
-  private static final Map<String, WeakReference<LottieComposition>> WEAK_REF_CACHE =
+  private static final SparseArray<LottieComposition> STRONG_RAW_REF_CACHE = new SparseArray<>();
```
gpeal: This should be `RAW_RES_STRONG_REF_CACHE` (strong goes with ref not raw)
Same for the asset cache.

cyrilmottier:
I've fixed the naming of the caching `Map`s to both `RAW_RES_[WEAK|STRONG]_REF_CACHE` and `ASSET_[WEAK|STRONG]_REF_CACHE`

cyrilmottier:
Poke @gpeal

No news on this for the past 2 weeks. Is there something wrong? :)

gpeal:
Thanks @cyrilmottier! I'll push out a new update soon :)

cyrilmottier:
Thanks @gpeal. Do you have a ETA for the new version push?

gpeal:
@cyrilmottier It's on 2.3.0-SNAPSHOT now

cyrilmottier:
Yep I saw that but can't find it :). Has it been pushed to jCenter or Maven Central?

PR #525, opened 2017-11-15
Add setRepeatMode API

fabionuno:
## Change Summary

This PR allows you to define what an animation should do when it reaches the end. This setting is applied only when the loop is enabled.

The supported values are: 
`LottieDrawable.RESTART` - The animation restarts from the beginning.
`LottieDrawable.REVERSE` - The animation plays backward (and then forward again).
The default behavior is to restart the animation.

You can use the new `lottie_repeatMode` XML attribute to set this value too


gpeal:
@fabionuno Can you deprecate the current `loop(boolean)` method, have that call through to your method, and add the same loop count functionality?

fabionuno:
@gpeal Doing this change, what is the better strategy to disable the loop in your opinion?
Can we create our own enum {None, Restart, Reverse } in LottieAnimationView class and use it as argument to setLoopMode?

fabionuno:
@gpeal  I've updated the code

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -594,8 +596,35 @@ public void removeAnimatorListener(Animator.AnimatorListener listener) {
     lottieDrawable.removeAnimatorListener(listener);
   }
 
+  /**
+   * @see #setLoopMode(int)
+   */
+  @Deprecated
   public void loop(boolean loop) {
-    lottieDrawable.loop(loop);
+    lottieDrawable.setLoopMode( loop ? LottieDrawable.RESTART : LottieDrawable.NONE );
```
gpeal: nit: extra spaces around predicate

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -63,6 +66,29 @@
   private int alpha = 255;
   private boolean performanceTrackingEnabled;
 
+
+  /** @hide */
```
gpeal: Why @hide this for a public API?

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -63,6 +66,29 @@
...
+  @IntDef({NONE, RESTART, REVERSE})
+  @Retention(RetentionPolicy.SOURCE)
+  public @interface LoopMode {}
+
+  /**
+   *  When the animation reaches the end and <code>loopMode</code> is NONE
```
gpeal: nit: extra spaces here

fabionuno:
@gpeal Wondering about the way that `setLoopMode` is working, adding the `setLoopCount` method we will have to define which method has priority to enable/disable the loop.

Does this animation have the loop enabled or not?
```xml
    ...
    app:lottie_fileName = "Logo / LogoSmall.json"
    app:lottie_loopMode = "restart"
    app:lottie_loopCount = "0" />
```

I suppose that `setLoopMode` should have the priority.
In this way, the previous sample has an infinite loop.

gpeal:
@fabionuno These APIs are more or less just pass throughs to ValueAnimator so I would expect your example to not loop (see [docs](https://developer.android.com/reference/android/animation/ValueAnimator.html#RESTART))

fabionuno:
@gpeal I've updated the code

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -528,10 +557,63 @@ public void setProgress(@FloatRange(from = 0f, to = 1f) float progress) {
...
+   * setting is applied only when loop is enabled
+   * Defaults to {@link #NONE}.
+   *
+   * @param loopMode {@link #NONE}, {@link #RESTART} or {@link #REVERSE}
+   */
+  public void setLoopMode(@LoopMode int loopMode) {
+    if (loopMode == NONE) {
+      animator.setRepeatCount(0);
+    } else {
+      if (animator.getRepeatCount() == 0) animator.setRepeatCount(ValueAnimator.INFINITE);
```
gpeal: @fabionuno Why do you have to modify repeat count here? Setting the mode should be sufficient. In other words, it can be NONE with a non-zero repeat count. 
This should mirror the [behavior of ValueAnimator](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/animation/ValueAnimator.java#879) as much as possible.

Also, single statement ifs should still have braces.

fabionuno: @gpel To start a loop with ValueAnimator we should use `setRepeatCount` and `setRepeatMode`. And to disable the loop we should call `setRepeatCount` with zero.

But I think most people use simple loops with animation restarting infinitely.
I've tried to create an easier way, just setting the `setLoopMode(Restart)` without the need to explicitly use `setLoopCount(Infinite)`.

We can make Lottie works exactly like `ValueAnimator` or we can keep the current proposal, saving the  LoopCount value when LoopMode is set to None and restore it if `LoopMode` is changed.

What do you think is the best way?

gpeal: @fabionuno I think this should mirror ValueAnimator as closely as possible meaning they can basically be pass through APIs to LottieValueAnimator itself. Since we added repeatCount as part of this PR, we can remove NONE and have the APIs be properly straight passed through to ValueAnimator. In that case, it would make sense to rename the methods from loop to repeat since REVERSE isn't really a loop technically

fabionuno:
@gpeal I've fixed the code according to the review

LottieSample/src/main/res/layout/fragment_list.xml
```diff
@@ -17,7 +17,7 @@
             android:layout_height="wrap_content"
             android:layout_gravity="center"
             app:lottie_fileName="Logo/LogoSmall.json"
-            app:lottie_loop="true" />
+            app:lottie_repeatCount="-1" />
```
gpeal: @fabionuno Hmm, this API is nice and consistent with the code now but it doesn't feel as nice to use. I'd be fine with leaving loop=true in (as you left in view_holder_grid below) as a convenience for this.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -63,6 +66,26 @@
...
 
+  @IntDef({RESTART, REVERSE})
+  @Retention(RetentionPolicy.SOURCE)
+  public @interface RepeatMode {}
+
+  /**
+   * When the animation reaches the end and <code>repeatCount</code> is INFINITE
+   * or a positive value, the animation restarts from the beginning.
+   */
+  public static final int RESTART = 1;
```
gpeal: Can you set these values directly equal to ValueAnimator.RESTART, etc?

gpeal: `public static final int RESTART = ValueAnimator.RESTART;`

gpeal:
@fabionuno This is super close! Thanks for putting in this work and all of these revisions. Looking forward to getting this in.

fabionuno:
@gpeal Thanks for your tips!

PR #543, opened 2017-11-29
Fix bugs:

LittleOrchid:
1. The layer does not disappear while other layer begins to appear. So the two layers are overlapping.
2. When computing the alpha of the layer, we should take into account the parentLayer of it.

Before opening a pull request:

* If you added a new feature or fixed a bug, add a json file to assets/Tests and add a test 
for it in `LottieTest#testAll`.
* Run screenshot tests by running `./gradlew --daemon recordMode screenshotTests` on a Nexus 
5 5.1.0 emulator and commit all screenshots that have changed.
* Assign gpeal as a reviewer.


gpeal:
Closing this for now. Reopen if it's still relevant.

PR #570, opened 2017-12-18
Move density scaling from parsing to rendering

gpeal:
Keyframes will now have their original AE values rather than pre-scaled values. This should make debugging easier.


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-ed5c32739e610958347f0f0fcfb990a45a96f8a5-android26&a=4

Issue #579, opened 2018-01-02
Make LottieDrawable implement the Animatable interface


alexjlockwood:
This would make it compatible with existing code that only understands how to work with Animatable objects.

gpeal:
@alexjlockwood Would you like to do a PR? :)

alexjlockwood:
Here's a pull request. :) #580

PR #580, opened 2018-01-03
make lottie drawable implement animatable

alexjlockwood:
This pull request updates `LottieDrawable` to implement the `Animatable` interface. See #579.

The implementation behaves the same as it does for `AnimatedVectorDrawable`s: https://github.com/aosp-mirror/platform_frameworks_base/blob/oreo-mr1-release/graphics/java/android/graphics/drawable/AnimatedVectorDrawable.java#L893-L945

TLDR:
* `start()` calls `playAnimation()`
* `stop()` calls `animator.end()`
* `isRunning()` calls `isAnimating()`

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
@@ -9,6 +9,7 @@ import android.graphics.ColorFilter;
 import android.graphics.Matrix;
 import android.graphics.PixelFormat;
 import android.graphics.Typeface;
+import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
 import android.os.Build;
 import android.support.annotation.FloatRange;
@@ -45,7 +46,8 @@ import java.util.Set;
  * handles bitmap recycling and asynchronous loading
  * of compositions.
  */
-@SuppressWarnings({"WeakerAccess", "unused"}) public class LottieDrawable extends Drawable implements Drawable.Callback {
+@SuppressWarnings({"WeakerAccess", "unused"})
+public class LottieDrawable extends Drawable implements Drawable.Callback, Animatable {
   private static final String TAG = LottieDrawable.class.getSimpleName();
 
   private interface LazyCompositionTask {
@@ -302,6 +304,19 @@ import java.util.Set;
   }
 
 // <editor-fold desc="animator">
+
+  @Override public void start() {
+    playAnimation();
+  }
+
+  @Override public void stop() {
+    lazyCompositionTasks.clear();
+    animator.end();
+  }
+
+  @Override public boolean isRunning() {
+    return isAnimating();
+  }
 
   /**
    * Plays the animation from the beginning. If speed is < 0, it will start at the end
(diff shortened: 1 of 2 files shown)


alexjlockwood:
One thing I wasn't sure about w/ the implementation for `stop()`...

Would it cause problems if `stop()` was called before the lazy composition tasks completed? Not sure if it will cause problems... since the animator will be given its end value in that case.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -303,6 +305,19 @@ public void clearComposition() {
 
 // <editor-fold desc="animator">
 
+  @Override public void start() {
+    playAnimation();
+  }
+
+  @Override public void stop() {
+    lazyCompositionTasks.clear();
```
gpeal: @alexjlockwood If you call stop then play, does AVD resume from where stop was called or restart from 0?

alexjlockwood: Calling `stop()` then `start()` on an AVD will restart the animation completely (it doesn't support pausing/resuming).

AVD source code: https://github.com/aosp-mirror/platform_frameworks_base/blob/oreo-mr1-release/graphics/java/android/graphics/drawable/AnimatedVectorDrawable.java#L893-L945

alexjlockwood: The `Animatable` docs don't specify how this method should be implemented exactly... so it is probably fine to do it however you want (i.e. `stop()` could call `cancelAnimation()` or `pauseAnimation()` or something).

gpeal: @alexjlockwood let's mimic AVD. Just call cancelAnimation

alexjlockwood: We can do that, but that doesn't exactly mimic AVD. :)

As you can see from the [source code](https://github.com/aosp-mirror/platform_frameworks_base/blob/oreo-mr1-release/graphics/java/android/graphics/drawable/AnimatedVectorDrawable.java#L893-L945), AVD implements `Animatable#stop()` by calling `animator.end()`, not `animator.cancel()`.

gpeal: @alexjlockwood Good point. I've never directly exposed end. Maybe we should do that. What do you think?

alexjlockwood: You mean by creating a `stopAnimation()` on `LottieDrawable` and `LottieValueAnimator` (and then have the `LottieValueAnimator` call its own `end()` method)?

gpeal: @alexjlockwood Expose `endAnimation` that proxies to `end()` then have `stop()` call `endAnimation()`
What do you think?

alexjlockwood:
Done!

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
alexjlockwood: Updated the pull request with your suggestion. LMK what you think!

PR #598, opened 2018-01-16
Fix rotations on solid layers

### LINKED ISSUE #587: Animation works on iOS but looks strange on Android
Hi,
This animation (attached) works perfectly on iOS, but behaves really strangely on Android.
Could you guys possibly give us an idea as to why?
[coffeanim.zip](https://github.com/airbnb/lottie-android/files/1616141/coffeanim.zip)

Kind Regards,
Dustyn

gpeal:
Prior to this, the transform matrix got mapped to a solid layer's bounds. However, if there is a rotation then the final shape won't be a rect so we need to map the points instead.

Fixes #587


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-65858e94727165f80cd205df650d7c3291763c58-android26&a=4

PR #607, opened 2018-01-30
Treat intersect masks like add masks

### LINKED ISSUE #601: Animations glitchy on Android
There was an issue #558 which was similar problem to this one - animations work fine on iOS, but are glitchy on Android. In that issue, someone said issue was fixed in 2.3.2-SNAPSHOT but this version is not available anymore and the newest library version (2.5.0-beta3) still has some issues with some of the animations we use.

You can look at mentioned issue to see how view is defined and how animation is played. I've also added `app:lottie_enableMergePathsForKitKatAndAbove="true"` but it did not help.

I've attached zip with two animations we have problem with. If aep files are needed, I can provide those too.
[Animations.zip](https://github.com/airbnb/lottie-android/files/1653923/Animations.zip)

gpeal:
This is not always correct but can be very close and is at least
better than doing nothing.
Lottie will also log when this happens.
Fixes #601

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
@@ -11,6 +11,7 @@ import android.graphics.RectF;
 import android.support.annotation.CallSuper;
 import android.support.annotation.FloatRange;
 import android.support.annotation.Nullable;
+import android.util.Log;
 
 import com.airbnb.lottie.L;
 import com.airbnb.lottie.LottieComposition;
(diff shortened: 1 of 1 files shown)


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-9dc78de25b97a823984caa62dfcb02480e9d3414-android26&a=4

lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
```diff
@@ -320,16 +321,26 @@ private void intersectBoundsWithMatte(RectF rect, Matrix matrix) {
...
-      paint = subtractMaskPaint;
-    } else {
-      paint = addMaskPaint;
+    switch (maskMode) {
+      case MaskModeSubtract:
+        paint = subtractMaskPaint;
+        break;
+      case MaskModeIntersect:
+        Log.w(L.TAG, "Animation contains intersect masks. They are not supported but will be " +
+            "treated like intersect masks.");
```
mister11: It should say: "treated like add masks" :smile:

gpeal: @mister11 Thanks 😛

PR #619, opened 2018-02-08
rename confusing words, remove duplicate in gradle dependency

JSpiner:
### 1. changed duplicated gradle depencency.

```
  dependencies {
    classpath 'org.ajoberstar:grgit:1.9.3'
    classpath 'com.android.tools.build:gradle:3.0.1'
    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
    classpath "org.jetbrains.kotlin:kotlin-android-extensions:$kotlinVersion"
    classpath 'org.ajoberstar:grgit:1.9.3' // removed
  }
```

### 2. rename confusing words.

The `lottie_filename` in xml and` .setAnimation () `in java are the same function. 
But their names are so different.
Many people(include me!), will get confused every time they use this.
So, I changed name `lottie_filename` to `lottie_setAnimation`.

Now, we can use it like below. (looks very uniform !)
```xml
        android:id="@+id/yellow_color_filter"
        android:layout_width="50dp"
        android:layout_height="50dp"
        android:background="#ffffff"
        app:lottie_colorFilter="#ffff00"
        app:lottie_setAnimation="HamburgerArrow.json"/>
```

```java
LottieAnimationView animationView = (LottieAnimationView) findViewById(R.id.animation_view);
animationView.setAnimation("hello-world.json");
animationView.loop(true);
animationView.playAnimation();
```

### DIFF
--- a/lottie/src/main/res/values/attrs.xml
+++ b/lottie/src/main/res/values/attrs.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <declare-styleable name="LottieAnimationView">
-        <attr name="lottie_fileName" format="string" />
+        <attr name="lottie_setAnimation" format="string" />
         <attr name="lottie_rawRes" format="reference" />
         <attr name="lottie_autoPlay" format="boolean" />
         <attr name="lottie_loop" format="boolean" />
--- a/LottieSample/src/main/res/layout/activity_test_color_filter.xml
+++ b/LottieSample/src/main/res/layout/activity_test_color_filter.xml
@@ -12,7 +12,7 @@
         android:layout_height="50dp"
         android:background="#ffffff"
         app:lottie_colorFilter="#ffff00"
-        app:lottie_fileName="HamburgerArrow.json"/>
+        app:lottie_setAnimation="HamburgerArrow.json"/>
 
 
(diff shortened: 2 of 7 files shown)


JSpiner:
I think I don't have permission to add reviewers. 
Please review it! @gpeal

lottie/src/main/res/values/attrs.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <declare-styleable name="LottieAnimationView">
-        <attr name="lottie_fileName" format="string" />
+        <attr name="lottie_setAnimation" format="string" />
```
gpeal: We definitely can't break the existing API here. Hundreds of people use it. It would have to at least fall back on the new one.
I would consider `lottie_asset` to be consistent with `rawRes` though. If we wanted to make them more parallel, we could deprecate `setAnimation` in favor of `setAsset` as well.

JSpiner: @gpeal 
I agree with you 👍 
Can I make a new PR that following features?
- add deprecate annotation to `setAnimation()`
- add new method `setAsset()` or etc

gpeal:
I'm going to close this due to inactivity. Feel free to reopen it if you want to continue.

PR #624, opened 2018-02-12
Reimplemented LottieValueAnimator

### LINKED ISSUE #491: playAnimation(int, int) yields wrong start and end progress
It looks like when LottieDrawable converts a frame number to progress, it divides the frame by the total number of frames:
 - https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java#L407
 - https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java#L455
 - https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java#L471

This is causing the progress to be smaller than expected. The frame should be divided by one less than the total frame count in order to have a progress in [0, 1] inclusive.

### LINKED ISSUE #555: Setting setMinAndMaxProgress for a second time does nothing on versions below 8.0
Hi,

First of all thanks for a great library. 

The problem seems to happen on versions of Android below 8.0. I have an animation of a progress bar, I do cicles to animate from A to B until I run out of points.

This works fine on Android 8.0 and above but below it does not throws errors but the second iteration does nothing.

I call this again:

                levelBar.setProgress(progresoActual);
                levelBar.setMinAndMaxProgress(progresoActual, progresoFinal);

With new values, for example 0 and 0.3 and I can't see the animation at all. I make sure I also reset min and max from previous iteration to avoid crashes, so by default I put them on 0 and 1 before doing anything else.

All works well on my Nexus 5x but in any other device with older Android it does not.

Any help is more than welcome

### LINKED ISSUE #610: setMinAndMaxFrame may be off by one frame for the min value
i found that when i had to update the `playAnimation` call - previously we could pass a frame range, now to effect the same i needed to `setMinAndMaxFrame` before calling `playAnimation`, here i note that i was needing to add +1 to the min frame value. could this also need the same fix you had implemented for set/get frames?

Thanks,
Shyam

### LINKED ISSUE #615: Appintro Animation start progress is not the same on different devices
On testing my appIntro animation it works well on my Xiaomi Mi A1  Android Oreo, it starts on the expected progress which is 0.20f shown in [this screenshot](https://drive.google.com/open?id=1-ejQkg19QUGuD2lSytlTuOUAdHuBc0eg).
But on testing on a Samsung Galaxy S4 Android Lollipop 5.0.1 it starts on a different Progress shown in [this screenshot](https://drive.google.com/open?id=1FQHHv6EEwg9ALC5jbjo3lWNnh9B09Fmo), I don't know the reason for this I tried to Debug to check the progress value that the AnimationView takes on the start and it takes the correct value which is 0.20f.
Here is my AppIntro Class code 

```
public class AppIntroActivity extends IntroActivity {

    private static final float[] ANIMATION_TIMES = new float[]{

            0.20f,
            0.60f,
            1f,
            1f

    };
    private LottieAnimationView animationView;
    private LockableViewPager viewPager;

    @Override
    protected Collection<? extends Fragment> generatePages(Bundle savedInstanceState) {
        return new ArrayList<EmptyFragment>() {{
            add(EmptyFragment.newInstance());
            add(EmptyFragment.newInstance());
            add(EmptyFragment.newInstance());
            
        }};
    }

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        animationView = (LottieAnimationView) LayoutInflater.from(this).inflate(R.layout.app_intro_animation_view, getRootView(), false);
        viewPager = (LockableViewPager) findViewById(com.matthewtamlin.sliding_intro_screen_library.R.id.intro_activity_viewPager);
        getRootView().addView(animationView, 0);
        setViewPagerScroller();
        addPageChangeListener(new ViewPager.OnPageChangeListener() {
            @Override
            public void onPageScrolled(int position, float positionOffset, int positionOffsetPixels) {
                setAnimationProgress(position, positionOffset);
            }

            @Override
            public void onPageSelected(int position) {
            }

            @Override
            public void onPageScrollStateChanged(int state) {
            }
        });

    }

    @Override
    protected IntroButton.Behaviour generateFinalButtonBehaviour() {
        return new IntroButton.Behaviour() {
            @Override
            public void setActivity(IntroActivity activity) {
                startActivity(new Intent(AppIntroActivity.this, MainActivity.class));
                finish();
            }

            @Override
            public IntroActivity getActivity() {
                return null;
            }

            @Override
            public void run() {
            }

        };
    }

    @Override
    public boolean progressIndicatorAnimationsAreEnabled() {
        return true;
    }

    private void setAnimationProgress(int position, float positionOffset) {

        float startProgress = ANIMATION_TIMES[position];
        float endProgress = ANIMATION_TIMES[position + 1];
        float lerb = lerp(startProgress, endProgress, positionOffset);
        animationView.setProgress(lerb);


    }


    private float lerp(float startValue, float endValue, float f) {
        return startValue + f * (endValue - startValue);
    }

    public static final class EmptyFragment extends Fragment {


        private static EmptyFragment newInstance() {
            return new EmptyFragment();

           
        }

        @Override
        public void onCreate(@Nullable Bundle savedInstanceState) {
            super.onCreate(savedInstanceState);


        }

        @Nullable
        @Override
        public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {

            return inflater.inflate(R.layout.fragment_empty, container, false);


        }
    }

    private void setViewPagerScroller() {
        //noinspection TryWithIdenticalCatches
        try {
            Field scrollerField = ViewPager.class.getDeclaredField("mScroller");
            scrollerField.setAccessible(true);
            Field interpolator = ViewPager.class.getDeclaredField("sInterpolator");
            interpolator.setAccessible(true);

            Scroller scroller = new Scroller(this, (Interpolator) interpolator.get(null)) {
                @Override
                public void startScroll(int startX, int startY, int dx, int dy, int duration) {
                    super.startScroll(startX, startY, dx, dy, duration * 7);
                }
            };
            scrollerField.set(viewPager, scroller);
        } catch (NoSuchFieldException e) {
            // Do nothing.
        } catch (IllegalAccessException e) {
            // Do nothing.
        }
    }}


```

and this my AnimationView XML Code 

```
    xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    android:id="@+id/animation_view"
    android:background="#EDEDED"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    app:lottie_fileName="appIntro.json"/>
```
And my appIntro.json File is attahced here 
[appIntro.json.zip](https://github.com/airbnb/lottie-android/files/1701555/appIntro.json.zip)

gpeal:
This version of LottieValueAnimator replaces all of the animator looping logic with a self-managed Choreographer callback architecture. There are two primary reasons for the change:
1) The conversion from setting frames to setting the currentPlayTime of the value animator often led to tricky floating point and rounding issues (#491, #610)
2) The ValueAnimator APIs and implementation have changed in different platform versions. For example, in some versions, the animator will report as running in the update callbacks and for some, it won't. (#555, #615)

There are other benefits as well:
1) Animations can now run independent of the animator duration scale setting.
2) Frame rates are now respected. Animations will now render at the fps specified in After Effects rather than 60fps.

@gshysun
Fixes #491 
Fixes #610 
Fixes #555 
Fixes #615


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-9ec55042b38380257f034d1cb9ff57e343b44fad-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-9ec55042b38380257f034d1cb9ff57e343b44fad-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-c2011194d52f839816dd253cc2855aa9d986ce25-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-87dfae7debb7a86be3139a83657032a60703d055-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-baf546485aa31ac6cde7dc89eb32df5cd7d44a13-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-8a004162fc73d8d9b42b2e15d784ead1910ae259-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-3d058dada912bea8d5ac204dbff2acb1ee2ce95e-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-a0b8740231c2c8541253b8e446ee84be28cce0c5-android26&a=4

rosconabn:
From this pull request I see new changes , what build can I use for the issues found in 2.5.0  related to frame animation to be slow ? rc-1 was fine already. Should I continue using that ?

gpeal:
@rosconabn This change was made between rc1 and rc2

PR #627, opened 2018-02-15
Get version name from gradle.properties

MizzleDK:
As titled. When doing future releases, there should now be one place less to update the `versionName` attribute because `build.gradle` will now fetch the attribute directly from `gradle.properties`.

If there's a better way of doing this, I'm all ears.

Tests are green:

lottie-tests

### DIFF
--- a/lottie/build.gradle
+++ b/lottie/build.gradle
@@ -10,7 +10,7 @@ android {
     minSdkVersion 16
     targetSdkVersion 27
     versionCode 100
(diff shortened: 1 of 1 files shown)


gpeal:
Thanks! Mind updating this too?
https://github.com/airbnb/lottie-android/blob/master/version.sh

gpeal:
@MizzleDK bump

PR #675, opened 2018-03-11
Remove null restriction on value callbacks

### LINKED ISSUE #638: Can't remove a color filter.
Pre `2.5` i was using `addColorFilterToLayer("LayerName", filter)` to add a color filter and cleared it by passing `null`.

In 2.5 I can't do that any longer. I can use `LottieValueCallback#setValue(null)`, but then it crashes at runtime with:.
`You must provide a static value in the constructor , call setValue, or override getValue.`

Also I think the case that a value must be set, should be documented. 
Also I don't see the use of allowing a `null` and then crashing at runtime. For a clearer API it should be strictly `@NotNull` and throw an `IllegalArugmentException immediately when setting null.

gpeal:
This prevented you from clearing a ColorFilter
Fixes #638

### DIFF
--- a/LottieSample/src/androidTest/java/com/airbnb/lottie/LottieSnapshotProvider.java
+++ b/LottieSample/src/androidTest/java/com/airbnb/lottie/LottieSnapshotProvider.java
@@ -434,6 +434,12 @@ public class LottieSnapshotProvider extends SnapshotProvider {
         new LottieValueCallback<ColorFilter>(new SimpleColorFilter(Color.GREEN)));
 
     testDynamicProperty(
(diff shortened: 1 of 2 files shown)


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-dc1fba70c20b2b40a30b17ec5cb3181a3822c856-android26&a=4

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-ce62be8c01f373e594e48c4e86470aa3f5551e7a-android26&a=4

PR #691, opened 2018-03-17
DEFAULT_CACHE_STRATEGY exposed.

dryganets:
This is needed in order to make more sophisticated and efficient update for the react-native plugin for Android.


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -50,8 +50,12 @@
  * {@link R.attr#lottie_progress}
  */
 @SuppressWarnings({"unused", "WeakerAccess"}) public class LottieAnimationView extends AppCompatImageView {
+
+  public static final CacheStrategy DEFAULT_CACHE_STRATEGY = CacheStrategy.Weak;
```
dryganets: Need it to use in react-native plugin change.

PR #694, opened 2018-03-19
Fix race condition with min/max frame

Kevinrob:
Composition can be loaded after we have set a custom `minFrame` and/or `maxFrame`.
The composition will override custom values.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
@@ -121,7 +121,11 @@ public class LottieValueAnimator extends BaseLottieAnimator implements Choreogra
 
   public void setComposition(LottieComposition composition) {
     this.composition = composition;
-    setMinAndMaxFrames((int) composition.getStartFrame(), (int) composition.getEndFrame());
+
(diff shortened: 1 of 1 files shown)


Kevinrob:
@gpeal 👋 Ping for a review 😄

lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
```diff
@@ -121,7 +121,22 @@ private float getFrameDurationNs() {
 
   public void setComposition(LottieComposition composition) {
     this.composition = composition;
-    setMinAndMaxFrames((int) composition.getStartFrame(), (int) composition.getEndFrame());
+
+    int minFrame;
+    if (this.minFrame != Integer.MIN_VALUE) {
```
gpeal: @Kevinrob Can't you just use the Math.min and math.max case for both? If the minFrame or maxFrame are MIN_VALUE or MAX_VALUE then it should use the start frame or end frame so the result would be the same as the else branch right?

Kevinrob: You are right! I changed it 👍

PR #729, opened 2018-04-13
Fix rendering graph in other themes

hearsilent:
## Before
device-2018-04-13-133128
device-2018-04-13-133146

## After
device-2018-04-13-132937
device-2018-04-13-133724

### DIFF
--- a/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/PlayerFragment.kt
+++ b/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/PlayerFragment.kt
@@ -63,6 +63,7 @@ private class UiState(private val callback: () -> Unit) {
     var speed by BooleanProperty(false)
     var trim by BooleanProperty(false)
 }
+
 class PlayerFragment : Fragment() {
 
     private val transition = AutoTransition().apply { duration = 175 }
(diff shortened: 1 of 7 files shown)


hearsilent:
@gpeal Please review this, thx 💯

LottieSample/src/main/kotlin/com/airbnb/lottie/samples/PlayerFragment.kt
```diff
@@ -335,6 +340,17 @@ class PlayerFragment : Fragment() {
         updateUiFromState()
     }
 
+    private fun invertColor(color: Int) {
+        val shouldInvertColor = getContrastColor(color) == Color.WHITE
+        animationView.isActivated = shouldInvertColor
+        toolbar.isActivated = shouldInvertColor
+    }
+
+    private fun getContrastColor(color: Int): Int {
```
gpeal: You can make this a private extension function like:
```
@ColorInt
private fun Int.contrastColor(): Int { ... }
```

gpeal: It's also a bit strange that you return a color but then just check if the color is white to create shouldInvertColor.

You could rename it `Int.requiresLightForeground()` or `Int.isDark()` or something

hearsilent:
@gpeal I benefited greatly from your suggestion.
I had updated this function, you can check it 👍

PR #749, opened 2018-05-04
Add basic support for base64 URLs.

stristr:
See [corresponding PR](https://github.com/airbnb/lottie-ios/pull/622) for `lottie-ios`.

Please refer to the new example animation spec, `WeAcceptInlineImage.json`, for reference syntax (it should render identically to the existing `WeAccept.json`). I tested this locally using a different asset with no noticeable runtime lag, but this approach should definitely be scrutinized for possible performance issues.

As noted in `lottie-ios` PR, please note that (due to the simple way images are handled in the browser) this works already in `lottie-web`.


lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
```diff
@@ -62,37 +63,54 @@ public void setDelegate(@Nullable ImageAssetDelegate assetDelegate) {
...
+      byte[] data;
       try {
-        if (TextUtils.isEmpty(imagesFolder)) {
-          throw new IllegalStateException("You must set an images folder before loading an image." +
-              " Set it with LottieComposition#setImagesFolder or LottieDrawable#setImagesFolder");
-        }
-        is = context.getAssets().open(imagesFolder + imageAsset.getFileName());
-      } catch (IOException e) {
-        Log.w(L.TAG, "Unable to open asset.", e);
+        data = Base64.decode(filename.substring(filename.indexOf(',') + 1), Base64.DEFAULT);
```
gpeal: Does the encoding of the base64 string not matter? It looks like that part of the string is never used

stristr: Yeah, I wondered about possible edge cases here as well, but I believe it should be fine to skip. From my testing, it seems like the ability to resolve a file suffixes (e.g. `.jpg`) to a universally recognized MIME types (e.g. `image/jpeg`) isn't actually required to interpret the binary data of a bitmap resource. This makes some sense, since virtually every bitmap encoding has a [file signature](https://en.wikipedia.org/wiki/List_of_file_signatures). For example, with [PNG](http://www.libpng.org/pub/png/spec/1.2/PNG-Structure.html), the first 8 bytes (hexadecimal) are _always_ `89 50 4E 47 0D 0A 1A 0A`. And everything that knows how to display a PNG basically knows those 8 bytes identify a bitmap image as having PNG encoding, and that the actual image data begins after the file signature.

Of note, the `BitmapFactory.Options` class has an [`outMimeType`](https://developer.android.com/reference/android/graphics/BitmapFactory.Options.html#outMimeType), but no corresponding `inMimeType`. I take that to be Android's way of saying MIME type will always be inferred if we know how to render this image at all.

I also tested what happens when you provide valid base64 data that isn't actually image data, and got a "good" result (no crash) :smile: using `data:image/foobar;base64,YmFy`.

bodymovin:
@stristr I've added support on the bodymovin plugin to export assets on the json encoded as base64. I hope it works well with this PR.

stristr:
Super cool @bodymovin! Just checking—did you use the same convention I adopted here, where the "directory" `"u"` is empty string (`""`), and the "filename" `"p"` is the data URI (`"data:image/…"`)?

bodymovin:
@stristr yes, but I've had to add an extra property "e" to let the player know the file is inline.
If not, whenever the json was not at the root level, it would try to prepend the path.
So it might break sometimes.

stristr:
Cool, thanks for the update @bodymovin! I'll check if any additional changes are needed in either this PR or the iOS one (not yet merged).

Issue #750, opened 2018-05-06
Repeat animation take the shortest time


EladJarby:
Hi,
First i set lootie for loop "true"`
i have a few animations that i set to same view.
For example, first animation duration is 4000.
After i cancel and clear animation and set the second animation with duration of 800 all working great,
But when i set first animation again (duration of 4000) "onAnimationRepeat" is executing every 800 and not 4000..
So i think that when we change animation it taking the shortest time and not the updating duration after setting the new animation to same view.

PR #752, opened 2018-05-07
Reset min/max frames unless no composition is set

gpeal:
This should handle both cases where:
1) You set a min/max frame while the composition is still loading #694 
2) You change the composition to one with a different length #750

Added a bunch of tests to confirm as well


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-ff1da4c671d4a57893602f7639487e775c31eb19-android26&a=4

PR #753, opened 2018-05-07
Properly register ColorFilter animations

### LINKED ISSUE #690: colorFilterAnimation progress are never updated.
For example, on FillContent, the colorFilterAnimation is created, stored to a field, the getValue method is called, but this animation is never updated, meaning that it's progress is never updated. Is this not supposed to be set (not supported by design)?

I'm trying to add a valueCallback with a SimpleLottieValueCallback (lambda), with a ColorFilter, which will use the ValueCallbackKeyframeAnimation, which calls getProgress, but it is always 0.

gpeal:
ColorFilter animations were not properly registered as animations so they wouldn't update the animation when changed.
Fixes #690


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-effe8c7005316fde77f72fde36e21e8b5904da47-android26&a=4

PR #764, opened 2018-05-11
Use lerp calculation to get the min and max frame.

### LINKED ISSUE #763: setMinAndMaxProgress uses a different method of calculating frame than setMinProgress and setMaxProgress
Currently `LottieDrawable#setMinAndMaxProgress` use the broken version of calculating the the frame which was fixed in https://github.com/airbnb/lottie-android/pull/754.

Update `LottieDrawable#setMinAndMaxProgress` to use lerp based method method.

benjamincassidy:
Fixes #763

Added test for bug in LottieDrawableTest. Fixed tests for `setMinProgress` and `setMaxProgress` since the tests would pass if the with the broken method of calculating frame.

For example, given progress of 0.5, endFrame of 200, and startFrame of 0
`0.5 * (200 - 0) == 0 + (0.5 * (200 - 0))`

whereas given progress of 0.5, endFrame of 200, and startFrame of 20
`0.5 * (200 - 20) = 90`
`20 + (0.5 * (200 - 20) = 110`

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
@@ -449,10 +449,9 @@ public class LottieDrawable extends Drawable implements Drawable.Callback, Anima
       });
       return;
     }
-    setMinAndMaxFrame(
-        (int) (minProgress * composition.getDurationFrames()),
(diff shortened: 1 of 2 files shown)


benjamincassidy:
@gpeal ping for review

PR #766, opened 2018-05-12
Fixed end frame calculation

gpeal:
The end frame in the JSON represents the first instant after the animation ends. For example, if the end frame is 60, it should be visible until 59.999999... This fixes the logic for how end frames are calculated.


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-348f11702a6365d650d60eaf85415c9fa97668bb-android26&a=27

gpeal:
This is the important diff
It's a single composition with the loop happening from 0 to 50 Percent and the completion is from 50 to 100.

PR #790, opened 2018-06-05
change gradle-wrapper version to 4.4, as it adopt to android-gradle-plugin-3.1.2

leleliu008:
this project use android-gradle-plugin-3.1.2, it not support gradle 4.6 or above version yet, if we use  gradle 4.6 or above version, it report a error as follow:

Configuration on demand is not supported by the current version of the Android Gradle plugin since you are using Gradle version 4.6 or abov    e. Suggestion: disable configuration on demand by setting org.gradle.configureondemand=false in your gradle.properties file or use a Gradle version less than 4.6.

resolve this problem have at least two methods:
1、use a Gradle version less than 4.6.
2、setting org.gradle.configureondemand=false in gradle.properties file

configure on demand may speed up builds, and Gradle4.4 is enough, so use a Gradle4.4 is best choice.

### DIFF
--- a/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieApplication.kt
+++ b/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieApplication.kt
@@ -1,15 +1,25 @@
 package com.airbnb.lottie.samples
 
 import android.support.multidex.MultiDexApplication
+import android.util.Log
 import com.google.gson.FieldNamingPolicy
 import com.google.gson.GsonBuilder
-import com.jakewharton.retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory
 import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.Retrofit
+import retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory
 import retrofit2.converter.gson.GsonConverterFactory
+import java.util.concurrent.TimeUnit
 
 class LottieApplication : MultiDexApplication() {
-    val okHttpClient by lazy { OkHttpClient.Builder().build() }
+    // print HTTP protocol content
+    private val loggingInterceptor = HttpLoggingInterceptor(HttpLoggingInterceptor.Logger { Log.i("Lottie", it) }).apply { level = HttpLoggingInterceptor.Level.BODY }
+
+    val okHttpClient by lazy {
+        OkHttpClient.Builder()
+            .addInterceptor(loggingInterceptor)
+            .connectTimeout(30, TimeUnit.SECONDS)
+            .build() }
 
     val gson by lazy {
         GsonBuilder()
--- a/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieApplication.kt
+++ b/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieApplication.kt
@@ -1,23 +1,17 @@
 package com.airbnb.lottie.samples
 
 import android.support.multidex.MultiDexApplication
-import android.util.Log
 import com.google.gson.FieldNamingPolicy
 import com.google.gson.GsonBuilder
 import okhttp3.OkHttpClient
-import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.Retrofit
 import retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory
 import retrofit2.converter.gson.GsonConverterFactory
 import java.util.concurrent.TimeUnit
 
 class LottieApplication : MultiDexApplication() {
-    // print HTTP protocol content
-    private val loggingInterceptor = HttpLoggingInterceptor(HttpLoggingInterceptor.Logger { Log.i("Lottie", it) }).apply { level = HttpLoggingInterceptor.Level.BODY }
-
     val okHttpClient by lazy {
         OkHttpClient.Builder()
-            .addInterceptor(loggingInterceptor)
             .connectTimeout(30, TimeUnit.SECONDS)
             .build() }
 
--- a/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottiefilesViewModel.kt
+++ b/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottiefilesViewModel.kt
@@ -53,6 +53,7 @@ class LottiefilesViewModel(application: Application) : AndroidViewModel(applicat
         disposables.add(observable
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
+                .retry(3)
                 .doOnSubscribe { loading.value = true }
                 .subscribe({
                     responses.add(it)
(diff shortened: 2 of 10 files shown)


LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieApplication.kt
```diff
@@ -1,15 +1,25 @@
...
 
 class LottieApplication : MultiDexApplication() {
-    val okHttpClient by lazy { OkHttpClient.Builder().build() }
+    // print HTTP protocol content
+    private val loggingInterceptor = HttpLoggingInterceptor(HttpLoggingInterceptor.Logger { Log.i("Lottie", it) }).apply { level = HttpLoggingInterceptor.Level.BODY }
+
+    val okHttpClient by lazy {
+        OkHttpClient.Builder()
+            .addInterceptor(loggingInterceptor)
+            .connectTimeout(30, TimeUnit.SECONDS)
```
leleliu008: I set a connectTimeout, because some country visit [https://www.lottiefiles.com](https://www.lottiefiles.com) spent a lot of  time.

LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottiefilesViewModel.kt
```diff
@@ -53,6 +53,7 @@ class LottiefilesViewModel(application: Application) : AndroidViewModel(applicat
         disposables.add(observable
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
+                .retry(3)
```
leleliu008: some country visit [https://www.lottiefiles.com](https://www.lottiefiles.com) spent a lot of time, and often occur SocketTimeoutException, so I want retry 3 times when occur exception.

LottieSample/src/main/kotlin/com/airbnb/lottie/samples/ShowcaseViewModel.kt
```diff
@@ -24,6 +24,7 @@ class ShowcaseViewModel(application: Application) : AndroidViewModel(application
         disposables.add(lottiefilesService.getCollection(COLLECTION)
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
+                .retry(3)
```
leleliu008: some country visit [https://www.lottiefiles.com](https://www.lottiefiles.com) spent a lot of time, and often occur SocketTimeoutException, so I want retry 3 times when occur exception.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
leleliu008: this project use android-gradle-plugin-3.1.2, it not support gradle 4.6 or above version yet, if we use gradle 4.6 or above version, it report a error as follow:

Configuration on demand is not supported by the current version of the Android Gradle plugin since you are using Gradle version 4.6 or abov e. Suggestion: disable configuration on demand by setting org.gradle.configureondemand=false in your gradle.properties file or use a Gradle version less than 4.6.

resolve this problem have at least two methods:
1、use a Gradle version less than 4.6.
2、setting org.gradle.configureondemand=false in gradle.properties file

configure on demand may speed up builds, and Gradle4.4 is enough, so use a Gradle4.4 is best choice.

lottie/src/main/java/com/airbnb/lottie/parser/AsyncCompositionLoader.java
```diff
@@ -21,7 +22,8 @@
     try {
       return LottieComposition.Factory.fromJsonSync(params[0]);
     } catch (IOException e) {
-      throw new IllegalStateException(e);
```
leleliu008: here throw a exception will case brash! If  params[0] is a invalid JSON file, it will cause crash, but user can not catch the exception.

lottie/src/main/java/com/airbnb/lottie/parser/AsyncCompositionLoader.java
```diff
@@ -21,7 +22,8 @@
...
+      L.error("", e);
+      return null;
```
leleliu008: the comment of `void onCompositionLoaded(@Nullable LottieComposition composition);` say _Composition will be null if there was an error loading it. Check logcat for more details._ I fix it according to this comment.

LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieApplication.kt
```diff
@@ -1,15 +1,25 @@
...
+import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.Retrofit
+import retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory
 import retrofit2.converter.gson.GsonConverterFactory
+import java.util.concurrent.TimeUnit
 
 class LottieApplication : MultiDexApplication() {
-    val okHttpClient by lazy { OkHttpClient.Builder().build() }
+    // print HTTP protocol content
+    private val loggingInterceptor = HttpLoggingInterceptor(HttpLoggingInterceptor.Logger { Log.i("Lottie", it) }).apply { level = HttpLoggingInterceptor.Level.BODY }
```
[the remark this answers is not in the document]
gpeal: @leleliu008 Can you remove this? I don't want to log all network requests.

gradle/wrapper/gradle-wrapper.properties
gpeal: It is best to stay on the latest version of all tools. It looks like it was deprecated because it isn't actually necessary with newer versions. Please disable it and upgrade gradle instead.
https://github.com/gradle/gradle/issues/2868

lottie/src/main/java/com/airbnb/lottie/parser/AsyncCompositionLoader.java
gpeal: This is a major change. I will introduce a new API with proper error handling.

gpeal: Please revert it.

lottie/src/main/java/com/airbnb/lottie/parser/AsyncCompositionLoader.java
gpeal: This is a breaking API change. Please revert it from this PR. I have it on my roadmap to improve the error handling for loading.

leleliu008: ok, I will revert it.

gradle/wrapper/gradle-wrapper.properties
leleliu008: I will revert the version of gradle to 4.7

LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieApplication.kt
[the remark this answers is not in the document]
leleliu008: I will remove it.

PR #808, opened 2018-06-16
Change repositories order to fix gradle sync

Litun:
Fixed the issue [stackoverflow](https://stackoverflow.com/questions/50570134/could-not-find-common-jar-android-arch-corecommon1-1-0)


Litun:
Sorry, duplicate

Issue #721, opened 2018-04-09
Request: in sample app, allow to see time instead of frames


AndroidDeveloperLB:
Have a toggle or something, to show time instead of frames, including how long exactly is the animation duration.

I'm talking about this:

![image](https://user-images.githubusercontent.com/5357526/38498058-e3525a48-3c0b-11e8-98a4-385ca2d1b21c.png)

Frames aren't enough to know how long the animation takes. It can be 60FPS, and can be 2FPS...
Maybe it's a good idea to show the FPS too :)

AndroidDeveloperLB:
Why close it?

markstonehouse:
@AndroidDeveloperLB because it has been resolved now as per #814

AndroidDeveloperLB:
We have it now?
Cool!

PR #814, opened 2018-06-21
Added total frames as well as current and total seconds of animation as per request #721

markstonehouse:
ezgif  #com-video-to-gif


markstonehouse:
@gpeal

LottieSample/src/main/kotlin/com/airbnb/lottie/samples/PlayerFragment.kt
```diff
@@ -522,4 +525,22 @@ class PlayerFragment : Fragment() {
             }
         }
     }
+
+    /* Calculate and returns current and total frames along wih current and total seconds */
+    private fun updateFramesAndDurationLabel(animation: LottieAnimationView): String {
+
+        var label = ""
+
+        // current frame out of total frames
```
gpeal: @markstonehouse 
I think this method is confusing to read as is. I would suggest:
1) Most of the comments are pretty redundant. This code should be simple enough 
that it is self-documenting.
2) I would suggest something like this:
```
val currentFrame = ...
val totalFrames = ...
val progress = ...
return "$currentFrame/$totalFrames\n$progress%"
```

I think it would read more clearly.

markstonehouse:
@gpeal Updated as per feedback

LottieSample/src/main/kotlin/com/airbnb/lottie/samples/PlayerFragment.kt
```diff
@@ -522,4 +525,19 @@ class PlayerFragment : Fragment() {
             }
         }
     }
+
+    private fun updateFramesAndDurationLabel(animation: LottieAnimationView): String {
+        val currentFrame = animation.frame.toString()
+        val totalFrames = ("%.0f").format(animation.maxFrame)
+
+        val animationSpeed:Float = Math.abs(animation.speed)
```
gpeal: Nit: you're missing a space

gpeal:
@markstonehouse

markstonehouse:
@gpeal Added white space as per request

PR #820, opened 2018-06-26
Fix setMinAndMaxFrames

Kevinrob:
`public void setMinAndMaxFrames(int minFrame, int maxFrame)` started to behave strange when I updated. So I debug it. I found something very (very) strange.

If `composition` is `null`, `compositionMinFrame` will be set to `Float.MIN_VALUE`.  
And if I pass to the function `minFrame = 0` then... (attach your belt...)

`MiscUtils.clamp(minFrame, compositionMinFrame, compositionMaxFrame)`
=> `MiscUtils.clamp(0, Float.MIN_VALUE, Float.MAX_VALUE)`
**give `1.4E-45` (ouch!)**

... I don't know exactly why but `(Float.MIN_VALUE >= 0) == true`!
It's probably because the float are cast to `int`.

Anyway, elsewhere we use `Integer.MIN_VALUE` and `Integer.MAX_VALUE` for `this.minFrame` and `this.maxFrame` 👍 

capture

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
@@ -162,7 +162,7 @@ public class LottieValueAnimator extends BaseLottieAnimator implements Choreogra
   }
 
   public void setMinAndMaxFrames(int minFrame, int maxFrame) {
-    float compositionMinFrame = composition == null ? Float.MIN_VALUE : composition.getStartFrame();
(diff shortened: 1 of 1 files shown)


Kevinrob:
@gpeal 👋 Can you review my PR please 😃

gpeal:
@Kevinrob Oh wow, this is interesting. Per [this](https://stackoverflow.com/a/11420190/715633) SO post, I think the correct thing to do is to use `-Float.MAX_VALUE`

Kevinrob:
@gpeal I just read the SO post and... 😯! So, you are right, here, it's `-Float.MAX_VALUE` that we want!

gpeal:
Naming is hard 😛

PR #828, opened 2018-07-06
Fix ImageLayer.java may use a recycled bitmap 

YongdongHe:
Use a recycled bitmap may produce a crash

image


YongdongHe:
@gpeal hi，could you review this pull request?

PR #841, opened 2018-07-16
Add an intersect mask warning at parse time

### LINKED ISSUE #784: It's said "Animation contains intersect masks", which does not exist (android)
So I'm currently making the loading images for the new project.
But my developer told me the warning message saying "W/LOTTIE: Animation contains intersect masks. They are not supported but will be treated like add masks." keeps on appearing, and I found the information that android OS doesn't support the intersect masks unlike iOS so I try to avoid the "intersect masks" as possible by separating the same logo layers with masks into 2 like the screen capture above (I changed the color of the eclipse mask to make it recognizable). 
So, as far as I can see, there is no intersect mask here, as it looks the same but each of the masks are divided.
BUT THAT WARNING MESSAGE STILL APPPEARS!
Is there no way to work with it besides: 
draw them frame by frame back again from the illustrator->get the shapes at the After Effects->attach the shapes frame by frames again and so on, just for the loading icon ??

The loading icon should be like this:
loading

Thank you.

gpeal:
Fixes #784


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-d5fa4184c449a176ad61ad915db3169056a61a63-android26&a=27

PR #859, opened 2018-07-26
Checking shape name nullability to avoid NullPointerException at KeyPath

anthonyfr:
Checking if shape key received is null to avoid a NullPointerException.


gpeal:
@anthonyfr I would like to avoid making Haiku specific changes to Lottie without understanding the full implications first. If could could reach out to them first, that would be great!

PR #926, opened 2018-09-05
add new attr in order to have the hierarchy keypaths

dlamande:
add new attr in order to have the hierarchy keypaths like iOS with the logHierarchyKeypaths methode.

Usage :  airbnb:lottie_logHierarchyKeypaths="true"


gpeal:
@dlamande I think it would make more sense to add a `compositionLoadedListener` so you could listen for that and do whatever you want. I could see other uses for that like playing a video or syncing multiple animations. 

This attr would make it very easy for people to accidentally leave logs in production which we don't want either.

dlamande:
@gpeal great :) i will do it ! thanks for feedback

PR #928, opened 2018-09-05
Add on composition loaded listener

dlamande:
new listener in order to know when composition is load and do whatever you want.

* see usage in DynamicActivity with KeyPathsHierarchyLogge

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -154,6 +154,10 @@ import java.util.Set;
           LottieDrawable.INFINITE));
     }
 
+    if (ta.getBoolean(R.styleable.LottieAnimationView_lottie_logHierarchyKeypaths, false)) {
+      lottieDrawable.setLogHierarchyKeypaths(true);
+    }
+
     setImageAssetsFolder(ta.getString(R.styleable.LottieAnimationView_lottie_imageAssetsFolder));
     setProgress(ta.getFloat(R.styleable.LottieAnimationView_lottie_progress, 0));
     enableMergePathsForKitKatAndAbove(ta.getBoolean(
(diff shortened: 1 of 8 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -504,6 +506,13 @@ public void setComposition(@NonNull LottieComposition composition) {
     setImageDrawable(lottieDrawable);
 
     requestLayout();
+
+    for (LottieOnCompositionLoadedListener lottieOnCompositionLoadedListener : lottieOnCompositionLoadedListeners) {
+      if(lottieOnCompositionLoadedListener != null) {
```
gpeal: This should never happen. `addCompositionLoadedListener` isn't marked as `@Nullable`. If you want to make it extra safe, you can do a preconditions check in `addLottieOnCompositionLoadedListener` but don't just silently ignore nulls here

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -891,6 +900,18 @@ private void enableOrDisableHardwareLayer() {
     setLayerType(useHardwareLayer ? LAYER_TYPE_HARDWARE : LAYER_TYPE_SOFTWARE, null);
   }
 
+  public void addLottieOnCompositionLoadedListener(LottieOnCompositionLoadedListener lottieOnCompositionLoadedListener) {
+    this.lottieOnCompositionLoadedListeners.add(lottieOnCompositionLoadedListener);
```
gpeal: the `this` is redundant in all 3 methods here.

gpeal: Also, add and remove return boolean. You might as well return them.

dlamande:
@gpeal review considered by adding `@NonNull` annotation, I think it's the better way for users.

PR #959, opened 2018-09-24
LottieCompositionFactory and Caching Cleanup

### LINKED ISSUE #958: RawRes caching uses two different key names
Was trying to figure out how to warm the composition cache and it seemed to not be working correctly for some reason and went digging through and I think the cache is being filled with the same composition twice when loading, since `fromJsonReaderSync(JsonReader reader, String cacheKey)` puts the composition with the key `"rawRes_" + resId` and then `LottieAnimationView.setAnimation(int rawRes)` also puts the composition but this time the key is just the `resId` as a string.

```java
public class LottieCompositionCache {
  […]
  @Nullable
  public LottieComposition getRawRes(@RawRes int rawRes) {
    return get(Integer.toString(rawRes));
  }
  […]
  public void put(@RawRes int rawRes, @Nullable LottieComposition composition) {
    put(Integer.toString(rawRes), composition);
  }
  […]
}
```

```java
public class LottieCompositionFactory {
  […]
  public static LottieResult<LottieComposition> fromRawResSync(Context context, @RawRes int rawRes) {
    try {
      return fromJsonInputStreamSync(context.getResources().openRawResource(rawRes), rawResCacheKey(rawRes));
    } catch (Resources.NotFoundException e) {
      return new LottieResult<>(e);
    }
  }

  private static String rawResCacheKey(@RawRes int resId) {
    return "rawRes_" + resId;
  }
  […]
}
```

Here's a screenshot of the cache data after showing one image, preloading two and showing one of those two:

There are duplicates with different keys. I think you sound change `rawResCacheKey` to behave the same as `LottieCompositionCache.put()`.

gpeal:
This cleans up a number of things related to the LottieCompositionFactory.
It:
* Handled task caching at the LottieCompositionFactory level so it is guaranteed to work everywhere. The existing architecture left several instances where compositions would not be cached.
* Simplified the way tasks are handled in `LottieAnimationView`
* Fixes #958 in which different cache keys are used for rawRes animations
* Removed deprecated factory APIs

### DIFF
--- a/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieFontViewGroup.kt
+++ b/LottieSample/src/main/kotlin/com/airbnb/lottie/samples/LottieFontViewGroup.kt
@@ -19,7 +19,6 @@ import java.util.*
 class LottieFontViewGroup @JvmOverloads constructor(
         context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
 ) : FrameLayout(context, attrs, defStyleAttr) {
-    private val compositionMap = HashMap<String, LottieComposition>()
     private val views = ArrayList<View>()
 
     private val cursorView: LottieAnimationView by lazy { LottieAnimationView(context) }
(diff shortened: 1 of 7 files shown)


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-c5c92a58306eead1f1bf7fca2a8b4a5c9b9d7487-android26&a=27

gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-8c2b497355cdfb9d641a109f614cc7e3d0776c9c-android26&a=27

PR #960, opened 2018-09-24
Allow returning null from value animator callbacks

### LINKED ISSUE #879: Listen to layer (object) position
Is there any way to listen to an object position that is part of the lottie animation? I can change its position by applying a value callback to its keypath. What I need is to listen to its position without modifying it.

If not, will you handle it in future releases?

I tried to use LottieValueCallback to get the position:

``` 
lottieAnimationView.addValueCallback(
                    new KeyPath("Gift Outlines"),
                    LottieProperty.TRANSFORM_POSITION,
                    new LottieValueCallback<PointF>(){
                        @Override
                        public PointF getValue(LottieFrameInfo<PointF> frameInfo) {
                            return super.getValue(frameInfo);
                        }
                    }
            );
```

The problem is that `super.getValue(frameInfo)`  always  returns null. I assumed `super.getValue(frameInfo)` return the position of the "Gift Outlines" at the current call and it will be called each time "Gift Outlines" moves . I need t to listen to "Gift Outlines" position.

gpeal:
Fixes #879

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/animation/keyframe/ColorKeyframeAnimation.java
+++ b/lottie/src/main/java/com/airbnb/lottie/animation/keyframe/ColorKeyframeAnimation.java
@@ -1,7 +1,7 @@
 package com.airbnb.lottie.animation.keyframe;
 
-import com.airbnb.lottie.value.Keyframe;
 import com.airbnb.lottie.utils.GammaEvaluator;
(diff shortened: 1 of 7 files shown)


gpeal:
Snapshot Test Results: https://happo.io/compare?q=lottie-master-android26..lottie-072157e3c63f854a398901b3db7709eaf4bd0c95-android26&a=27

PR #971, opened 2018-10-08
Added possibility to modify standard task listeners

iismagilov:
Default failure listener throws exception and there is no way to catch it because callback is posted to Handler.  
Two new setters are added in this PR  in order to configure listeners for success and failure cases.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -54,13 +54,15 @@ import java.util.Set;
 
   private static final String TAG = LottieAnimationView.class.getSimpleName();
 
-  private final LottieListener<LottieComposition> loadedListener = new LottieListener<LottieComposition>() {
+  @NonNull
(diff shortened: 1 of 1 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -375,6 +377,38 @@ public void setAnimationFromUrl(String url) {
...
   }
 
+  /**
+   * Sets a task listener. Previous listener will be removed
+   *
+   * @param loadedListener - task listener
+   * @see LottieTask#addListener(LottieListener)
+   * @see LottieTask#removeListener(LottieListener)
+   */
+  public void setLoadedListener(@NonNull LottieListener<LottieComposition> loadedListener) {
```
gpeal: `add`/`remove` would be better than `set`/`remove` (so you need to keep track of a list)

If the task changes after you set the listener, it also won't add your listener

gpeal: @iismagilov

gpeal:
Closing this because it is stale. Feel free to reopen it if it is needed.

Issue #978, opened 2018-10-19
LottieCompositionFactory.fromUrl  has no  Memory cache of  composition 


yjiang2009:
can you give me some suggest?  i use lottieview.setAnimationFromUrl( ) to load animation from net(is zip,contains json and images)，i saw the source has use  disk cache 。but  i want first use  memory cache ,then use disk cache .last  from net .if
i just use setAnimationFromUrl  or  LottieCompositionFactory.fromUrl，it there have two same url  ,they have no memory cache ,they will load from net once ,then load from disk ,but i want load from memory the second time to keep lottiew  fast 。
if  i  use lottiecomposition disk cache  myself .i dont know the image folder default ,they are load direct from zip by  fromZipStreamSync.,  how can i solve this ? thank you .

gpeal:
This is fixed as of that PR.

PR #986, opened 2018-10-31
Added memory cache layer for async network requests

RhysDavis:
Added a call to the LRU Memory cache for lottie network requests.

I've also removed the asyncFetch calls from the NetworkFetcher since the url lottie composition creates an Async task from the LRU cache call

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -52,8 +52,13 @@ public class LottieCompositionFactory {
    * future use. Because of this, you may call `fromUrl` ahead of time to warm the cache if you think you
    * might need an animation in the future.
    */
-  public static LottieTask<LottieComposition> fromUrl(Context context, String url) {
-    return NetworkFetcher.fetch(context, url);
+  public static LottieTask<LottieComposition> fromUrl(final Context context, final String url) {
+    String urlCacheKey = "url_" + url;
+    return cache(urlCacheKey, new Callable<LottieResult<LottieComposition>>() {
+      @Override public LottieResult<LottieComposition> call() {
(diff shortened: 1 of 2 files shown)


gpeal:
@RhysDavis Can you test this?

elihart:
Is this new code any different from what `fromZipStream` does? if that works with bitmaps and caching then it seems loading from url would be the same.

Can we also update `LottieAnimationView#setAnimationFromUrl` to use `LottieCompositionCache`, similar to what `setAnimation` does, so that we get an in memory cache for the url animation once it is loaded?

Gnod:
@elihart we can't use like that, the problem it quiet the same as I mention above. As the bitmap might be recycle when the view is detach from the view, ImageAssetManager#bitmapForId currently can't measure that we can get bitmap again.  As the setAnimation is use a resource with image from the imagesFolder,  the ImageAssetManager#bitmapForId has the logic to get the bitmap back from the imagesFolder.

RhysDavis:
@elihart `LottieAnimationView#setAnimationFromUrl` does go through the composite cache, I don't pass a `cacheKey` since I generate one based on the URL that's being loaded.  ie `LottieAnimationView#setAnimationFromUrl` calls `LottieCompositeFactory#fromUrl` directly which will hit the animation cache.

@Gnod if what you're saying is true, wouldn't the cache implementation be broken for all methods of loading a lottie file? The cache implementation isn't exclusive for network urls

Gnod:
@RhysDavis the other logic is ok as the ImageAssetManager#bitmapForId method had logic to load the bitmap from imagesFolder we set. But if the image is from network, currently logic will not find the image and load it as a bitmap again if the origin bitmap had been recycled.

gpeal:
Fyi #1002

gpeal:
Fixes #978

PR #1013, opened 2018-11-27
Remove bitmap recycling

### LINKED ISSUE #1002: Remove bitmap recycling
it is no longer necessary on newer versions of Android.

gpeal:
Fixes #1002

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -159,25 +159,16 @@ import java.util.Set;
   }
 
   @Override public void setImageResource(int resId) {
-    recycleBitmaps();
     cancelLoaderTask();
     super.setImageResource(resId);
   }
 
   @Override public void setImageDrawable(Drawable drawable) {
-    setImageDrawable(drawable, true);
-  }
-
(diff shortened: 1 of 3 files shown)


azchohfi:
There are still references to "recycleBitmaps" on LottieDrawable:
https://github.com/airbnb/lottie-android/blob/d736d2180ccc25798a28cbb3c0db653950112d50/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java#L46

kotlinc:
@gpeal If possible, could you elaborate on the reason behind removing this recycleBitmaps() call?

Meanwhile, the doc is still referencing calling recycleBitmaps() when using LottieDrawable directly.
http://airbnb.io/lottie/#/android?id=srcassets

gpeal:
@kotlinc In modern versions of Android, GC is sufficient to release bitmap resources.

kotlinc:
Thank you for your reply @gpeal. The only reference I have collected so far is https://developer.android.com/topic/performance/graphics/manage-memory where it reads that `recycle()` is not recommended for API > 10 and `the pixel data is stored on the Dalvik heap along with the associated bitmap`. Do you happen to know other references that could confirm the behavior of GC's sufficiency for garbage collecting bitmap resources?

gpeal:
@kotlinc 
image

PR #1048, opened 2018-12-15
[Breaking Change] Automatic hardware acceleration detection

gpeal:
Dash path effects are the only feature that Lottie uses that isn't supported on all api levels. We should be able to safely detect this [one case for API < 28](https://developer.android.com/guide/topics/graphics/hardware-accel#unsupported) and turn on hardware acceleration otherwise.

This removes the existing APIs since they no longer have a function.


gpeal:
**Snapshot Tests**
Report: https://happo.io/a/27/report/39debd46980e011ccb54b5436f6bf49948674149-android26
Diff: https://happo.io/a/27/compare/master-android26/39debd46980e011ccb54b5436f6bf49948674149-android26

Issue #381, opened 2017-07-13
Hardware automatic recognition


talentqing:
Can you according to the features of animation to decide use hardware? Because most animations is so simple and no hardware porblem, if any animation use software will be stuck when the ui is complex and the mobile phone performance is bad.

for exmaple:
[test.zip](https://github.com/airbnb/lottie-android/files/1143984/test.zip)
this animation in hardware is performance obviously better than in software.

So I thing can according to the api of features and phone info to auto decide use hardware or software.

gpeal:
Yup, it's been on my mind to do something like this!

talentqing:
Do you plan to add this feature in the next release?

gpeal:
@talentqing Not sure, maybe.

gpeal:
@talentqing You can enable it manually today though.

AndroidDeveloperLB:
@gpeal But how could we know when it's safe to enable it?
According to what I've found (written about it [here](https://github.com/airbnb/lottie-android/issues/682)) , I think it's safe to use it starting from Android L (API 21)

gpeal:
I've decided to not do this because some animations perform much worse with hardware acceleration even when it supports it. It should continue to be used on a case by case basis.

gpeal:
http://airbnb.io/lottie/android/performance.html#hardware-acceleration

AndroidDeveloperLB:
@gpeal Can you please provide an example (animation file) that enabling hardware acceleration makes it perform worse?
Also, do you really thing the default behavior, of having it disabled - is usually performing better ?
If not, why having this as the default behavior?
After all, even choosing not to do anything still have a default behavior, which is having it disabled.
This was the case on Android too: for some time, hardware acceleration was disabled by default, and starting from certain Android version, it became enabled by default. Developers can still disable it if they want.

gpeal:
@AndroidDeveloperLB Please [read the article](http://blog.danlew.net/2015/10/20/using-hardware-layers-to-improve-animation-performance/) linked to in the docs. 
There are also cases where certain phones have GAIA bugs that cause things to render incorrectly with hardware acceleration even if they are supported. Facebook stopped using hardware acceleration to render reactions for this exact reason. That's why it's disabled by default. It's easy to enable though so feel free to use it if it works well for you.

AndroidDeveloperLB:
This article is very old, and I've read it already. In fact some of my comments are there.

Question is, if it's safe to use now starting from certain Android version, why not enable by default?
[...]

PR #1072, opened 2018-12-26
Add renderMode to control automatic/software/hardware acceleration

gpeal:
While testing, I discovered that animations with lots of/large masks and mattes perform *significantly* worse with hardware acceleration than software acceleration on pre-Pie devices because of RenderNode#textureUpload. It is really hard to detect this dynamically so I'm leaving the default on automatic (with an additional heuristic for >4 mattes and masks) but having an option to set it to hardware/software manually.

#381


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/3cd7bfc9fc525f7f0dd9ef0e421d79240144d9f2-android28) [Diff](https://happo.io/a/27/compare/master-android26/3cd7bfc9fc525f7f0dd9ef0e421d79240144d9f2-android28)

PR #1097, opened 2019-02-02
Fix null cacheKeys Returning Cached Tasks

### LINKED ISSUE #1092: Multiple successive calls to LottieCompositionFactory.cache() with a null cacheKey return the same LottieTask
This is making multiple successive calls to LottieComposition.Factory methods for different input streams return the same composition, preventing easy upgrade.

gpeal:
Previously, setting a null cache key could return a cached task or result. This explicitly prevents that from happening. The new tests failed before and pass now.
Fixes #1092


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/7d62a3cfba805b603927efee844cc06404776aef-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/7d62a3cfba805b603927efee844cc06404776aef-android28)

PR #1098, opened 2019-02-02
Pause Lottie when visibility changes

### LINKED ISSUE #987: Looping animations continue when LottieAnimationView is invisible
I recently did some benchmarking on an app which uses lottie to do some animations (autoplay and looping). I noticed that there is quite some CPU usage when the app is in the background and tried to investigate.

It seems to me looping animations do not pause/stop when the containing `LottieAnimationView` is off screen, and/or the Activity is paused. 

I believe this is due to the cleanup code being only in `onDetachedFromWindow()` which is not necessarily being called once the Activity goes into a paused state and most definitely not, when the view is simply not visible (`GONE`, `INVISIBLE` ) anymore.

I have created a little test project illustrating the issue:
https://github.com/florianPOLARSTEPS/lottie_leak 

Starting it up you will see an empty screen, with an invisible (`GONE`) lottie view, that shows a toast each time the animation is repeated. If you put the app in the background, the toast will continue.

gpeal:
onAttachedToWindow and onDetachedFromWindow aren't sufficient for LottieAnimationView to properly pause and resume. We must do the same from onVisibilityChanged as well or we risk Lottie animating when the app is in the background.

Fixes #987 
Thanks @florianPOLARSTEPS

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -21,6 +21,7 @@ import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.util.JsonReader;
 import android.util.Log;
+import android.view.View;
 
 import com.airbnb.lottie.model.KeyPath;
 import com.airbnb.lottie.value.LottieFrameInfo;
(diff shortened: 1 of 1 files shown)


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/44b9e52398c00b27cc224e436630b4c75b51f5de-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/44b9e52398c00b27cc224e436630b4c75b51f5de-android28)

PR #1115, opened 2019-02-15
Fix adjustment scaling for text alignment when drawing text with fonts.

codedjentleman:
We don't need to account for scaling when drawing fonts, because the scaled text size is already set as the text size on the paint.


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/d4f7d4460750bf5012ea383d9797fdae2e62653d-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/d4f7d4460750bf5012ea383d9797fdae2e62653d-android28)

gpeal:
**[Snapshots Failed](https://app.bitrise.io/build/b6fc66a1b6b20c06)**

PR #1125, opened 2019-02-18
Fix animations starting paused when autoPlay is set to true

### LINKED ISSUE #1123: Loader doesn't animate when switching to lottiefiles fragment

gpeal:
The assumed logic for the original code was that the `else` branch is for `visibility != VISIBLE` but that isn't true if `visibility == VISIBLE` and wasAnimatingWhenVisibilityChanged is false. That caused animations to incorrectly be paused the first time they became visible when autoPlay was true.

Fixes #1123


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/da8138e28c0cee335b95cc7d99454e65bac7f7ec-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/da8138e28c0cee335b95cc7d99454e65bac7f7ec-android28)

PR #1174, opened 2019-04-05
Fix issues about visibility changes

Dong-wan:
#### 1) The animation could be resumed although it is not visible.
`onVisibilityChanged` callback will be called even when the visibilities of the ancestors of the view has changed.
So we need to check the visibility with `isShown()`, not with `visibility == View.VISIBLE` in onVisibilityChanged callback.
https://developer.android.com/reference/android/view/View#isShown()

**ex)** Assume below view hierarchy when the activity is in background.
DecorView (invisible)
...
|- FrameLayout (invisible)
|- |- LottieAnimationView (visible)

When the activity goes foreground, onVisibilityChanged callback in LottieAnimationView will be called by DecorView's visibility changing.
```java
onVisibilityChanged(DecorView /* changedView */, true /* visibility */)
```
Then, LottieAnimationView will resume animation although it is not visible by FrameLayout.
DecorView (visible)
.. 
|- FrameLayout (invisible)
|- |- LottieAnimationView (visible)

#### 2) `playAnimation()` and `resumeAnimation()` could be called while not visible.
`playAnimation()` and `resumeAnimation()` should be blocked if the view is not visible,
and the animation should be resumed when the view goes visible.
So, we need to set `wasAnimatingWhenNotShown = true` in `playAnimation()` and `resumeAnimation()` when the view is not visible.

#### 3) `pauseAnimation()` and `cancelAnimation()` were not worked when the activity is in background.
The animation could be resumed by `onVisibilityChanged` callback.
We need to reset `wasAnimatingWhenNotShown = false` in `pauseAnimation()` and `cancelAnimation()`

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -72,7 +72,7 @@ import java.util.Set;
   private final LottieDrawable lottieDrawable = new LottieDrawable();
   private String animationName;
   private @RawRes int animationResId;
-  private boolean wasAnimatingWhenVisibilityChanged = false;
+  private boolean wasAnimatingWhenNotShown = false;
   private boolean wasAnimatingWhenDetached = false;
   private boolean autoPlay = false;
   private RenderMode renderMode = RenderMode.AUTOMATIC;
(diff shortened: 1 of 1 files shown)


Dong-wan:
Also, #1173 will be fixed with this CL.

gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/41a8c31e437112720fd4c78e69122595cb510c30-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/41a8c31e437112720fd4c78e69122595cb510c30-android28)

gpeal:
Thanks!

liemvo:
The animation resume does not work properly on my app that has the Lottie animation view in the item of the recycler view. When I make a quick scroll while the Lottie view is animating then come back but the animation is not resumed.

PR #1176, opened 2019-04-09
change LottieAnimationView.failureListener' onResult callback into a protected method

AliMaster:
when lottie task failed, we can use the protected method "onLottieFailure" to do something like setting this view invisible and reporting to server what happens


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/86470c019c1ce200a06252ced0a22bf9c51db173-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/86470c019c1ce200a06252ced0a22bf9c51db173-android28)

AliMaster:
sure，that is a good way

PR #1187, opened 2019-04-17
No animation if systemAnimationScale == 0

CmoaToto:
If systemAnimationScale == 0, the lottie animation will play then immediately jump to the last frame, or only jump to the last frame if it is a loop animation. 
This solution handles if the dev move the animation at some percentage, change first or last frame or play the animation in reverse.

Really useful for testing purpose in order to not overload the main thread

Unittest are OK, I couldn't start the uiTest as it require the AWS key.

I couldn't add unittest as it obviously doesn't try to start the animations in it

There is a small Code Review here https://github.com/CmoaToto/lottie-android/pull/1

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -24,6 +24,7 @@ import android.util.Log;
 import android.view.View;
 
 import com.airbnb.lottie.model.KeyPath;
+import com.airbnb.lottie.utils.Utils;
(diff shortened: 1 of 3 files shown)


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/7fc88c2a72a2f0c3d3d194e70e73f5ec01d78599-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/7fc88c2a72a2f0c3d3d194e70e73f5ec01d78599-android28)

PR #1190, opened 2019-04-23
Have software rendering fallback on hardware acceleration

podarsmarty:
In cases where LottieAnimationView is large and software rendering is used, Android may not be able to generate a bitmap large enough which causes nothing to be shown. Rather than have this happen, fall back on hardware acceleration which is supported > api 11. Potentially taking a perf hit and displaying is better than a blank screen and not displaying at all.

Encountered this where upon rotation on a tablet, the view became big enough that generation of the bitmap failed.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -34,6 +34,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import static com.airbnb.lottie.RenderMode.HARDWARE;
+import static com.airbnb.lottie.RenderMode.SOFTWARE;
+
 /**
  * This view will load, deserialize, and display an After Effects animation exported with
  * bodymovin (https://github.com/bodymovin/bodymovin).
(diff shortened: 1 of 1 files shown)


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/4779c12c6f02d088361a264d36edd43cb09aafd4-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/4779c12c6f02d088361a264d36edd43cb09aafd4-android28)

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -806,6 +809,24 @@ private void clearComposition() {
     lottieDrawable.clearComposition();
   }
 
+  /**
+   * If rendering via software, Android will fail to generate a bitmap if the view is too large. Rather than displaying
+   * nothing, fallback on hardware acceleration which may incur a performance hit. minSdkVersion is 16. If this is ever
```
gpeal: @podarsmarty The minSdk won't be lowered so you can remove this comment but thanks for thinking of it 😄

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -806,6 +809,24 @@ private void clearComposition() {
...
+   *
+   * @see #setRenderMode(RenderMode)
+   * @see com.airbnb.lottie.LottieDrawable#draw(android.graphics.Canvas)
+   */
+  @Override
+  public void buildDrawingCache(boolean autoScale) {
+    super.buildDrawingCache(autoScale);
+    if (getLayerType() == LAYER_TYPE_SOFTWARE && getDrawingCache(autoScale) == null) {
+      setRenderMode(HARDWARE);
+      requestLayout();
```
gpeal: Why requestLayout()?

podarsmarty: hmm good point. i was thinking that since this is called during draw, we would need to retrigger it but the logic there would fall into hardware stuff since software is first. removing

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
podarsmarty: XD removed

gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/3a8b263e2d2ba2172f9b0ba41b7aaf77ec1238a3-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/3a8b263e2d2ba2172f9b0ba41b7aaf77ec1238a3-android28)

PR #1215, opened 2019-05-15
 Refer to the frameTimeNanos from Choreographer (#1214)

### LINKED ISSUE #1214: Lottie 3.0.1 is incompatible with Robolectric test
With Lottie 2.2.x, the unit test back by Robolectric works well with Lottie animation. But in 3.0.1, it starts to fail some test cases due to timeout(). Below is the root cause analysis:

The animation does be played for both version 2.2.x and 3.0.1 at unit test since the system animation scale is 1.0 instead of 0. And the way the LottieAnimation works in unit test is to cooperate with [ShadowLegacyChoreographer](https://github.com/robolectric/robolectric/blob/f9b5f5d4dc3b2b6acb8ec8614994569d0537bba9/shadows/framework/src/main/java/org/robolectric/shadows/ShadowLegacyChoreographer.java) to [accelerate](https://github.com/robolectric/robolectric/blob/f9b5f5d4dc3b2b6acb8ec8614994569d0537bba9/shadows/framework/src/main/java/org/robolectric/shadows/ShadowLegacyChoreographer.java#L153) the animation process without changing the test or app logic. And test case can control the pending/scheduled thread using APIs like: Robolectric.getForegroundThreadScheduler().advanceBy(delay, TimeUnit.MILLISECONDS).

The reasons 3.0.1 doesn't work is because:
In [LottieValueAnimator.java#doFrame()](https://github.com/airbnb/lottie-android/blob/3a4eacdac8ea921ed9b32d0e5dab18cf4b6d9d2a/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java#L85), it doesn't honor the **frameTimeNanos** injected by [ShadowLegacyChoreographer](https://github.com/robolectric/robolectric/blob/f9b5f5d4dc3b2b6acb8ec8614994569d0537bba9/shadows/framework/src/main/java/org/robolectric/shadows/ShadowLegacyChoreographer.java) (in Lottie 2.2.x, it did honor it because it inherits the frameCallback from [ValueAnimator.java](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/animation/ValueAnimator.java#1422)) so that the animation cannot be accelerated and lead to test case's timeout (some test cases wait the main thread never become idle to proceed). This can be fixed by using **frameTimeNanos** from **Choreographer** instead of using System.nanos() unless it was to fix some other issues. I tested the change locally and it doesn't seem to break anything.

boduan402:
This is to fix #1214 . Please check out the root cause analysis in #1214 .

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
@@ -82,7 +82,7 @@ public class LottieValueAnimator extends BaseLottieAnimator implements Choreogra
       return;
     }
 
-    long now = System.nanoTime();
+    long now = frameTimeNanos;
     long timeSinceFrame = now - lastFrameTimeNs;
     float frameDuration = getFrameDurationNs();
     float dFrames = timeSinceFrame / frameDuration;


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/2cd7cc417442f9f5fa91d8740e39505fccd0dc26-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/2cd7cc417442f9f5fa91d8740e39505fccd0dc26-android28)

PR #1222, opened 2019-05-19
Set correct frame when animation ends and speed is < 0

### LINKED ISSUE #1207: Frame is not at min frame when backward animation ends
When an animation is run with a negative speed and reaches the min frame thus ending the animation the frame value is then being set to the max frame value without consideration of whether the animation was being played backward. This does not reflect that actual displayed frame which is the min frame.

Looks like a check for animation direction probably needs to be added to LottieValueAnimator.java at line 99 for version 3.0.1

gpeal:
Fixes #1207


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/60aaf410f554a842e6b7dc7a5ddafe6f089ca316-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/60aaf410f554a842e6b7dc7a5ddafe6f089ca316-android28)

Issue #1241, opened 2019-05-30
Abnormal stop of lottie animation


zxc120301945:
Switch the fragment in the activity, then switch the fragment through the viewpager in the fragment. At this time, execute the lottie animation in one of the fragments, then switch the fragment through the activity, then jump to another activity, and then return to the previous activity. Then switch to the fragment that is executing the lottie animation, at this point the lottie animation stops.
View problems by code source
![image](https://user-images.githubusercontent.com/14074954/58607545-a127e480-82d2-11e9-8f9e-8445305181b0.png)
This method will be called when the fragment is switched, and then called again when the activity jumps. On the second call, the value of wasAnimatingWhenVisibilityChanged will be assigned false again by isAnimating(). This time, it will return to the currently executing lottie animation. In the fragment, the resumeAnimation() method in the onVisibilityChanged callback will not be executed.

gpeal:
I think this PR will fix it. Please try 3.0.4

zxc120301945:
OK

gpeal:
@zxc120301945 3.0.5 now

PR #1244, opened 2019-05-30
Fix old platform behavior which would prevent views from playing on attach

### LINKED ISSUE #1243: Regression: Animation does not play when playAnimation() called in RecyclerView onBindViewHolder
When setting an animation from a `RecyclerView.Adapter`'s `onBindViewHolder()` method, the animation will never play. This is a regression from 3.0.0 to 3.0.1, and I believe is caused by commit a483b7695e.

Essentially, when `onBindViewHolder()` is first called, the view is not attached to the window yet. With this change, the animation will not start but will set the `wasAnimatingWhenNotShown` flag to true. However, in `onAttachedToWindow()`, both the `autoPlay` and `wasAnimatingWhenNotShown` flags are checked to decide whether or not to play the animation. If the `lottie_autoPlay` attribute was not set in XML, then the animation will never start playback unless the `playAnimation()` call is deferred to after `onAttachedToWindow()`. 

I've attached a simple sample project here ([LottieBug.zip](https://github.com/airbnb/lottie-android/files/3237744/LottieBug.zip)) that reproduces the issue, but it can be reproduced as simply as the following:

```
class MainActivity : AppCompatActivity() {
  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)
    
    setContentView(
      RecyclerView(this).apply {
        layoutManager = LinearLayoutManager(this@MainActivity, LinearLayoutManager.VERTICAL, false)
        adapter = object : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
          override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): RecyclerView.ViewHolder {
            return object : RecyclerView.ViewHolder(LottieAnimationView(parent.context)) {}
          }

          override fun getItemCount(): Int = 1

          override fun onBindViewHolder(holder: RecyclerView.ViewHolder, position: Int) {
            (holder.itemView as LottieAnimationView).apply {
              loop(true)
              setAnimation("lottie.json")
              playAnimation()
            }
          }
        }
      }
    )
  }
}
```

gpeal:
The platform check fixes #1243 #1241 
The other logic makes `wasAnimatingWhenDetached` work even if autoPlay is false.

cc @kcoppock

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -259,8 +259,15 @@ import static com.airbnb.lottie.RenderMode.SOFTWARE;
 
   @Override protected void onAttachedToWindow() {
     super.onAttachedToWindow();
-    if (autoPlay && wasAnimatingWhenDetached) {
+    if (autoPlay || wasAnimatingWhenDetached) {
       playAnimation();
+      // Autoplay from xml should only apply once.
+      autoPlay = false;
+    }
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+      // This is needed to mimic newer platform behavior.
+      // https://stackoverflow.com/a/53625860/715633
+      onVisibilityChanged(this, getVisibility());
     }
   }
 


gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/ca0acf3f8aee920d6d10901c388096700b11c73a-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/ca0acf3f8aee920d6d10901c388096700b11c73a-android28)

Issue #1324, opened 2019-07-16
RecycleView, wasAnimatingWhenDetached isn't set to false when view is attached


YuriiTsap:
We have a related [issue](https://github.com/airbnb/lottie-android/issues/1322) created few hours ago.
But the main issue isn't highlighted/described enough inside of the ticket.

**_Let's start:_** 
I'm using `LottieAnimationView.java` inside `RecyclerView.java`. In my code every third item has an animation running, I'm operating with a dataset containing 60 items.
When `LottieAnimationView.java` is playing an animation and user starts scrolling  we receive a `onDetachedFromWindow()` (_in case if view scroll offset is big enough_).
```
@Override 
protected void onDetachedFromWindow() {
    if (isAnimating()) {
      cancelAnimation();
      wasAnimatingWhenDetached = true;
    }
    super.onDetachedFromWindow();
  }
```
Here we are setting `wasAnimatingWhenDetached` to `true`. Afterwards, when user scrolls again and returns to the same item - onAttachedToWindow() callback is called:
```
@Override protected void onAttachedToWindow() {
    super.onAttachedToWindow();
    if (autoPlay || wasAnimatingWhenDetached) {
      playAnimation();
      // Autoplay from xml should only apply once.
      autoPlay = false;
    }
    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
      // This is needed to mimic newer platform behavior.
      // https://stackoverflow.com/a/53625860/715633
      onVisibilityChanged(this, getVisibility());
    }
  }
```
In this callback we are relying on `wasAnimatingWhenDetached` and in our case we will play animation which wasn't finished. Until now everything was pretty ok, we are restarting animation which wasn't finished - pretty logical. But we aren't setting `wasAnimatingWhenDetached` flag to `false` - and that has some side effects. 
In example we scroll back and forth, basically just returning to the same item - and it will again go through `onDetachedFromWindow()` && `onAttachedToWindow()`callbacks. And when the `onAttachedToWindow()` callback will be triggered we will re-run the animation again, evenif it was finished, which isn't the desired behavior(_at least as I see it_). That will happen because `wasAnimatingWhenDetached` isn't set to false in `onAttachedToWindow()` callback. Here is an example how it works in real-life(almost) [app](https://drive.google.com/open?id=1j-2ct6f5mWlC8lO7bT1vq0svtjf9riM2). Video is attached as link on google drive as Github doesn't supports mp4.
[...]

PR #1340, opened 2019-08-01
Fixed endless recycler view animation. 

YuriiTsap:
Additionally - adjusted "testPausesWhenScrolledOffScreenAndResumesWhenComesBack" test.

See more details: https://github.com/airbnb/lottie-android/issues/1324

### DIFF
--- a/LottieSample/src/androidTest/java/com/airbnb/lottie/samples/FragmentVisibilityTests.kt
+++ b/LottieSample/src/androidTest/java/com/airbnb/lottie/samples/FragmentVisibilityTests.kt
@@ -15,24 +15,20 @@ import androidx.fragment.app.testing.launchFragmentInContainer
 import androidx.lifecycle.Lifecycle
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
-import androidx.test.espresso.Espresso
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.IdlingRegistry
 import androidx.test.espresso.action.ViewActions
-import androidx.test.espresso.assertion.ViewAssertions
 import androidx.test.espresso.assertion.ViewAssertions.matches
-import androidx.test.espresso.matcher.ViewMatchers
 import androidx.test.espresso.matcher.ViewMatchers.*
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
 import com.airbnb.lottie.LottieAnimationView
 import com.airbnb.lottie.LottieDrawable
-import com.airbnb.lottie.samples.R
 import com.airbnb.lottie.model.LottieCompositionCache
 import com.nhaarman.mockitokotlin2.mock
-import org.junit.Assert
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
+import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import kotlin.random.Random
@@ -41,7 +37,7 @@ import kotlin.random.Random
 @LargeTest
 class FragmentVisibilityTests {
 
-    @Test
+    @Before
     fun setup() {
         LottieCompositionCache.getInstance().clear()
     }
@@ -306,6 +302,8 @@ class FragmentVisibilityTests {
                 return RecyclerView(requireContext()).apply {
                     layoutManager = LinearLayoutManager(requireContext(), LinearLayoutManager.VERTICAL, false)
                     adapter = object : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+                        var animationWasPlayed = false
+
                         override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): RecyclerView.ViewHolder {
                             return when (viewType) {
                                 0 -> object : RecyclerView.ViewHolder(LottieAnimationView(parent.context).apply { id = R.id.animation_view }) {}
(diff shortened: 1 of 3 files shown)


LottieSample/src/androidTest/java/com/airbnb/lottie/samples/FragmentVisibilityTests.kt
```diff
@@ -41,7 +37,7 @@ import kotlin.random.Random
 @LargeTest
 class FragmentVisibilityTests {
 
-    @Test
+    @Before
```
YuriiTsap: Almost sure it should be a before setup :)

LottieSample/src/androidTest/java/com/airbnb/lottie/samples/FragmentVisibilityTests.kt
```diff
@@ -323,12 +321,17 @@ class FragmentVisibilityTests {
...
-                            animationView = holder.itemView as LottieAnimationView
-                            (holder.itemView as LottieAnimationView).apply {
-                                repeatCount = LottieDrawable.INFINITE
-                                setAnimation(R.raw.heart)
-                                playAnimation()
-                                IdlingRegistry.getInstance().register(LottieIdlingResource(this, name = "Lottie ${Random.nextFloat()}"))
+                            if (!animationWasPlayed) {
+                                animationView = holder.itemView as LottieAnimationView
+                                (holder.itemView as LottieAnimationView).apply {
+                                    //                                    repeatCount = LottieDrawable.INFINITE
```
YuriiTsap: In case of `LottieIdlingAnimationResource` we will wait forever if `repeatCount = LottieDrawable.INFINITE`. So I commented it out - in any case I can't see a reason to leave it here. It's not test critical condition.

gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/00617ffb18156a700c278b1be81d8b28770f8924-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/00617ffb18156a700c278b1be81d8b28770f8924-android28)

YuriiTsap:
> **Snapshot Tests**
> **28**: [Report](https://happo.io/a/27/report/00617ffb18156a700c278b1be81d8b28770f8924-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/00617ffb18156a700c278b1be81d8b28770f8924-android28)

Not sure what this comment means. Should I adjust something?

gpeal:
@YuriiTsap It's an automated step from CI

LottieSample/src/androidTest/java/com/airbnb/lottie/samples/FragmentVisibilityTests.kt
```diff
@@ -323,12 +321,17 @@ class FragmentVisibilityTests {
                         }
 
                         private fun bindLottieHolder(holder: RecyclerView.ViewHolder) {
-                            animationView = holder.itemView as LottieAnimationView
-                            (holder.itemView as LottieAnimationView).apply {
-                                repeatCount = LottieDrawable.INFINITE
-                                setAnimation(R.raw.heart)
-                                playAnimation()
-                                IdlingRegistry.getInstance().register(LottieIdlingResource(this, name = "Lottie ${Random.nextFloat()}"))
+                            if (!animationWasPlayed) {
```
gpeal: Why do you need this stateful ViewHolder?
If you want the animation to restart every time it is bound, you'd want to call play each time. If not, you can just call play in onCreate or set autoPlay to true in the xml

LottieSample/src/androidTest/java/com/airbnb/lottie/samples/LottieIdlingAnimationResource.kt
```diff
@@ -0,0 +1,37 @@
+package com.airbnb.lottie.samples
+
+import android.animation.Animator
+import androidx.test.espresso.IdlingRegistry
+import androidx.test.espresso.IdlingResource
+import com.airbnb.lottie.LottieAnimationView
+
+class LottieIdlingAnimationResource(animationView: LottieAnimationView?, private val name: String = "Lottie") : IdlingResource {
```
gpeal: I think the naming of this class is confusing. It's an idling resource that detects whether Lottie is animating. I think swapping Idling and Animation would be clearer.

LottieSample/src/androidTest/java/com/airbnb/lottie/samples/LottieIdlingAnimationResource.kt
```diff
@@ -0,0 +1,37 @@
...
+                isIdle = true
+                callback?.onTransitionToIdle()
+                animationView.removeAllAnimatorListeners()
+                IdlingRegistry.getInstance().unregister(this@LottieIdlingAnimationResource)
+            }
+        })
+    }
+
+    private var callback: IdlingResource.ResourceCallback? = null
+    private var isIdle = animationView?.isAnimating?.not() ?: true
```
gpeal: I think this should be:
`private val isIdle get() = animationView?.isAnimating?.not() ?: true`
Then yiou don't need to set it on lines 13 or 17

LottieSample/src/androidTest/java/com/airbnb/lottie/samples/FragmentVisibilityTests.kt
```diff
@@ -347,6 +350,75 @@ class FragmentVisibilityTests {
...
+    fun testPausesWhenScrolledOffScreenAndResumesWhenComesBackWithoutRepeatingWhenFinished() {
+
+        class TestFragment : Fragment() {
+            var animationView: LottieAnimationView? = null
+            override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+                return RecyclerView(requireContext()).apply {
+                    layoutManager = LinearLayoutManager(requireContext(), LinearLayoutManager.VERTICAL, false)
+                    adapter = object : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+
+                        var animationWasPlayed = false
```
gpeal: Same question here as above regarding the stateful ViewHolder

LottieSample/src/androidTest/java/com/airbnb/lottie/samples/LottieIdlingAnimationResource.kt
YuriiTsap: In such case we need to hold reference to `animationView: LottieAnimationView?`, make it val. We should decide between referencing a view directly and having a bit more code :) I don't really like referencing, but I'm almost sure it won't affect anything in current scenario.

LottieSample/src/androidTest/java/com/airbnb/lottie/samples/FragmentVisibilityTests.kt
YuriiTsap: Function `testPausesWhenScrolledOffScreenAndResumesWhenComesBack()` as I described in the original [issue comment](https://github.com/airbnb/lottie-android/issues/1324#issuecomment-514776627), from the declaration  - we are expecting that animation will continue playing, when we went back, by **scrolling up**, after stopping it in the middle by **scrolling down**. 

But in the scenario with **stateless** `ViewHolder` every time we **scroll up** `onBindViewHolder()`(of course if we've scrolled enough and our `ViewHolder` isn't valid anymore, in case of **10_000** scroll thats always true) callback is called and the `playAnimation()` is called also. So in any case, we aren't truly testing the  `LottieAnimation.onAttachedToWindow()` callback within its check:
```
if (autoPlay || wasAnimatingWhenDetached) {
      playAnimation();
//etc
}
```
We are just triggering another animation and testing:
` scenario.onFragment { assertTrue(it.animationView!!.isAnimating) }` 
Which doesn't make sense, as it will return `true` 100%. 

In case with **stateful** `ViewHolder` we aren't triggering animation more than one time, because of that we are verifying valid work of `LottieAnimation.onAttachedToWindow()` callback.

YuriiTsap:
> Thanks for looking in to this :)

Thanks for reviewing and merging.

PR #1349, opened 2019-08-10
Don't resume animation when system animation is disabled.

boduan402:
While the playAnimation() api call respects systemAnimationsEnabled flag, resumeAnimation() ignores it. Add the condition check to fix the case.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
@@ -406,7 +406,9 @@ public class LottieDrawable extends Drawable implements Drawable.Callback, Anima
       });
       return;
     }
-    animator.resumeAnimation();
+    if (systemAnimationsEnabled) {
+      animator.resumeAnimation();
+    }
   }
 
   /**


lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -406,7 +406,9 @@ public void run(LottieComposition composition) {
       });
       return;
     }
-    animator.resumeAnimation();
+    if (systemAnimationsEnabled) {
```
boduan402: In playAnimation() call, it checks 2 conditions like:
`if (systemAnimationsEnabled || getRepeatCount() == 0) {
      animator.playAnimation();
    }`
Do we need to do so here? Not get why need to check getRepeatCount() == 0 in playAnimation.

gpeal:
**Snapshot Tests**
**28**: [Report](https://happo.io/a/27/report/7e9893cf934c2f6878b9c050ac69ad9cc7e0613e-android28) [Diff](https://happo.io/a/27/p/27/compare/master-android28/7e9893cf934c2f6878b9c050ac69ad9cc7e0613e-android28)

PR #1361, opened 2019-08-20
Always use original context (Activity) when obtaining cache key and resource stream to correctly load and handle raw-night resources when the dark mode is on

### LINKED ISSUE #1305: How to work in Android Q dark theme
Android Q offers a new Dark theme.
Does LottieAnimationView support this behavior?
I put json on res/raw for light mode, res/raw-night for dark mode.
and in xml uses app:lottie_rawRes="@raw/sample".
I change theme to light or dark, it still plays the same json file.
I also use animationView.setAnimation(R.raw.sample), it still plays the same json file.
I'm pretty sure the json is changed(I dump the content) when theme for light to dark or dark to light mode.
so is anyone has the experience on this?
thanks.

plnice:
Application context isn't reliable source of information about dark mode (uiMode is UI_MODE_NIGHT_NO when the dark mode is on - Activity carries the correct information) and can incorrectly return default resource instead of the -night variant.

Fixes #1305

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -278,7 +278,9 @@ public class LottieCompositionFactory {
       com.airbnb.lottie.parser.moshi.JsonReader reader, @Nullable String cacheKey, boolean close) {
     try {
       LottieComposition composition = LottieCompositionMoshiParser.parse(reader);
-      LottieCompositionCache.getInstance().put(cacheKey, composition);
+      if (cacheKey != null) {
+        LottieCompositionCache.getInstance().put(cacheKey, composition);
+      }
       return new LottieResult<>(composition);
     } catch (Exception e) {
       return new LottieResult<>(e);
@@ -360,7 +362,9 @@ public class LottieCompositionFactory {
       }
     }
 
-    LottieCompositionCache.getInstance().put(cacheKey, composition);
+    if (cacheKey != null) {
+      LottieCompositionCache.getInstance().put(cacheKey, composition);
+    }
     return new LottieResult<>(composition);
   }
 
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -398,9 +398,6 @@ public class LottieCompositionFactory {
     task.addListener(new LottieListener<LottieComposition>() {
       @Override
       public void onResult(LottieComposition result) {
-        if (cacheKey != null) {
-          LottieCompositionCache.getInstance().put(cacheKey, result);
-        }
         taskCache.remove(cacheKey);
       }
     });
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -18,6 +18,7 @@ import org.json.JSONObject;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.ref.WeakReference;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


lottie/src/test/java/com/airbnb/lottie/LottieCompositionFactoryTest.java
```diff
@@ -83,11 +86,20 @@ public void testNonJsonAssetFile() {
...
+        LottieResult<LottieComposition> result = LottieCompositionFactory.fromRawResSync(contextRef, 0);
         assertNotNull(result.getException());
         assertNull(result.getValue());
     }
 
+    @Test
+    public void testContextNotReachable() {
+        WeakReference<Context> contextRef = new WeakReference<>(null);
+        LottieResult<LottieComposition> result = LottieCompositionFactory.fromRawResSync(contextRef, 0);
+        assertTrue(result.getException() instanceof NullPointerException);
```
gpeal: This behavior could introduce a regression since it will fail to parse in some cases where it would succeed today.
Can you make this hold the weak Activity and Application context and if Activity context is gone, default to day mode but continue to parse and cache it?

plnice: Sure! Forcepushed the change.

lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -146,8 +152,12 @@ public static void setMaxCacheSize(int size) {
    * The resource id will be used as a cache key so future usages won't parse the json again.
    */
   @WorkerThread
-  public static LottieResult<LottieComposition> fromRawResSync(Context context, @RawRes int rawRes) {
+  public static LottieResult<LottieComposition> fromRawResSync(WeakReference<Context> contextRef, Context appContext, @RawRes int rawRes) {
```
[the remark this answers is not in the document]
gpeal: You can add in the docs that you should use an Activity context where possible to get proper day night mode support but Lottie won't leak the Activity.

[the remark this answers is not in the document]
plnice: @gpeal pushed the changes; docs PR: https://github.com/airbnb/lottie/pull/97

lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -128,14 +129,19 @@ public static void setMaxCacheSize(int size) {
...
     // Prevent accidentally leaking an Activity.
+    final WeakReference<Context> contextRef = new WeakReference<>(context);
     final Context appContext = context.getApplicationContext();
     return cache(rawResCacheKey(context, rawRes), new Callable<LottieResult<LottieComposition>>() {
       @Override
       public LottieResult<LottieComposition> call() {
-        return fromRawResSync(appContext, rawRes);
+        @Nullable Context originalContext = contextRef.get();
+        Context context = originalContext != null ? originalContext : appContext;
+        return fromRawResSync(context, rawRes);
```
gpeal: If the activity is in night mode and then gets dereferenced and you fall back to the application context, it might report day mode and store the night asset there.

plnice: From what I observed, if the application context reports day mode (incorrectly), then the resources object from this context also returns the day asset (incorrectly). So in the case of Activity object not reachable, the flow is as follows:
1. The first call of `rawResCacheKey()` (the argument of the `cache` method) will return correct "night" key
1. In the `cache()` method, the callable is called, which obtains day resource from the application context and stores it under "day" cache key (re-generated from the application context in `fromRawResSync()`
1. Back in the `cache()` method, the callable result is stored in the cache under the "night" key - that's the only issue I can see here, I guess that the solution would be to change the callable signature from `Callable<LottieResult<LottieComposition>>` to something like `Callable<LottieResult<Pair<String /* new cache key* ,LottieComposition>>>` and store the result under new cache key passed in the `LottieResult`, but this would require changing the signature of `fromRawResSync()` method.

gpeal: @plnice Actually, I think we can remove the call where the `cache` function adds the composition to the cache. It is already done by `fromZipStreamSyncInternal` and `fromJsonReaderSyncInternal` so it is redundant there. If we do that then the `cache` function is just storing the task cache and to repro this bug, you would somehow have to join an in-progress task which seems highly improbable.

So the steps here would be:
1) Remove the the part of cache where it adds a composition to the cache
2) Wrap the existing cache.put calls in the 2 internal factory functions with `if (cacheKey != null)` (as it always should have been)

then I think this would be reasonable to land as-is.

plnice: Got it! Added two commits with the changes.

DazzyG:
Is there an ETA for this PR getting into a release?

Cheers

gpeal:
@plnice Thanks for working through this!

Issue #902, opened 2018-08-21
Animations with transparency transitions are rendered differently in Android


nima-ahmadi:
In a transparency transition (like a fade-out), Android seems to be adding up the overlapping layers (in the same group) while it is not happening in iOS/web. I think the iOS/web behaviour in expected.

[files.zip](https://github.com/airbnb/lottie-android/files/2306869/test.for.lottie.zip)

![lottie](https://user-images.githubusercontent.com/5301577/44410219-f2c86500-a531-11e8-8598-56551c82caa4.png)

gpeal:
This is a limitation of the way opacity works on Android. To achieve the correct behavior would incur a huge performance penalty unfortunately.

gpeal:
@nima-ahmadi  3.1.0 will have a new API `setApplyingOpacityToLayersEnabled` where you can manually enable this to render correctly. It isn't on by default because there is a performance overhead in doing so.

nima-ahmadi:
Great! Thanks for keeping Lottie awesome! @gpeal @r-ralph

Issue #606, opened 2018-01-29
Overlapping shapes with opacity keyframe fades have an artifact where overlapped


jeffsteinmetz:
Works on iOS
opacity fades show overlap on Android
(ignore the gradient background in this screen capture, the test was the overlapping circles)
<img width="264" alt="android_emulator_-_nexus_5_api_26_5554" src="https://user-images.githubusercontent.com/5488059/34169743-3aef4d7c-e49d-11e7-9f38-5b3ccae75f15.png">

Attaching .json and After Effects

This example uses a child composition with overlapping circles, and fades that comp in the main composition.

[fade-animation-example.zip](https://github.com/airbnb/lottie-react-native/files/1572584/fade-animation-example.zip)

gpeal:
Duplicate of #223

taylorling:
@gpeal I think I might have the same issue, and I am using pre-comp. The opacity change in the comp renders correctly in web and iOS, but in Android, it has this overlapping transparent effect.

Attached below is the comparison of iOS and Android:
![media-20180205](https://user-images.githubusercontent.com/7766387/35806994-58ec98b2-0abc-11e8-88c2-45d46e8befa2.png)
![img_0484](https://user-images.githubusercontent.com/7766387/35806995-59260b38-0abc-11e8-9eb1-42b445bdfe8c.PNG)

taylorling:
@gpeal Sorry for the follow up - should I file this as another issue, or this will not be fixed?

gpeal:
@taylorling This will not be fixed. The performance overhead would be too high. If you have ideas of efficient ways of solving it, let me know.

miguelnserra:
Same problem here. I usually use the pathfinder in Illustrator so separate the shapes or do it manually in After Effects.

gpeal:
@jeffsteinmetz @taylorling @miguelnserra  3.1.0 will have a new API `setApplyingOpacityToLayersEnabled` where you can manually enable this to render correctly. It isn't on by default because there is a performance overhead in doing so.

miguelnserra:
@gpeal Thanks!

Issue #223, opened 2017-03-26
Opacity applied to individual elements


podrivo:
Opacity is being applied to individual elements, so during the transition you can see each element, instead of seeing as a group.

So for instance, take this screenshot from After Effects of a hand being transitioned.

The hand is formed of different elements, that fade as group.
<img width="551" alt="screen shot 2017-03-26 at 1 14 28 pm" src="https://cloud.githubusercontent.com/assets/546221/24332966/2a38ee88-1226-11e7-9da4-4a314ce21ff7.png">

Same happens in Bodymovin:
<img width="1440" alt="screen shot 2017-03-26 at 1 15 49 pm" src="https://cloud.githubusercontent.com/assets/546221/24332976/674224c0-1226-11e7-86ee-9b524118b14a.png">

But this is the result from Lottie:
<img width="471" alt="screen shot 2017-03-26 at 1 23 16 pm" src="https://cloud.githubusercontent.com/assets/546221/24333029/6d6b0758-1227-11e7-8e1e-2561dc64575d.png">

Using beta3.
Here's the .aep and .json: [opacity.zip](https://github.com/airbnb/lottie-android/files/870878/opacity.zip)

gpeal:
Unfortunately, achieving this will incur a pretty heave performance penalty so I'd try alternative ways of achieving this. Precomps maybe?

gpeal:
@podrivo 3.1.0 will have a new API `setApplyingOpacityToLayersEnabled` where you can manually enable this to render correctly. It isn't on by default because there is a performance overhead in doing so.

PR #1362, opened 2019-08-22
Fix rendering issue of translucent object overlapping

r-ralph:
## Objective
As mentioned in #223, #606 and #902, there is a problem that when some translucent objects are drawn in the layer, they are not drawn as expected.
Fixes with off-screen rendering with minimal performance overhead.
In addition, make this feature toggleable to minimize the impact on the existing environment.

## ScreenShot

| Before | After |
| :-: | :-: |
| スクリーンショット 2019-08-22 18 21 37 | スクリーンショット 2019-08-22 18 21 43 |
| スクリーンショット 2019-08-22 18 27 30 | スクリーンショット 2019-08-22 18 27 38 |
| https://youtu.be/Mo40rMJuJh0 | https://youtu.be/ZONr1w4pSl4 |

I used the animation that is attached in #223


lottie/src/main/java/com/airbnb/lottie/animation/content/ContentGroup.java
```diff
@@ -148,25 +166,37 @@ Matrix getTransformationMatrix() {
...
+    boolean isRenderingWithOffScreen = isOffScreenRenderingEnabled && drawingContents.size() > 1 && layerAlpha != 255;
+    if (isRenderingWithOffScreen) {
+      offScreenRectF.set(0, 0, 0, 0);
+      getBounds(offScreenRectF, matrix, true);
+      offScreenPaint.setAlpha(layerAlpha);
+      Utils.saveLayerCompat(canvas, offScreenRectF, offScreenPaint);
+    }
+
+    int childAlpha = isRenderingWithOffScreen ? 255 : layerAlpha;
+    for (int i = drawingContents.size() - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
r-ralph: I don't investigate about that.
I've created the new list in order to get the size of `contents` with DrawingContent in L184. I think using the for statement twice was not good for performance.

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -920,6 +920,15 @@ public void setRenderMode(RenderMode renderMode) {
     enableOrDisableHardwareLayer();
   }
 
+  /**
+   * Sets whether to use an off-screen buffer to draw the layers contained in this composition.
+   * <p>
+   * Note: This process is very expensive and it's highly recommended that enable hardware acceleration if you want to turn on.
+   */
+  public void setOffScreenRenderingEnabled(boolean isOffScreenRenderingEnabled) {
```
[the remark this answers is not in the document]
r-ralph: Thanks for suggestion!

lottie/src/main/java/com/airbnb/lottie/animation/content/ContentGroup.java
```diff
@@ -148,25 +166,37 @@ Matrix getTransformationMatrix() {
     return path;
   }
 
-  @Override public void draw(Canvas canvas, Matrix parentMatrix, int parentAlpha) {
+  @Override public void draw(Canvas canvas, Matrix parentMatrix, int parentAlpha, boolean isOffScreenRenderingEnabled) {
```
gpeal: Pretty much all of these classes have access to LottieDrawable. I think you could just check it for `isOffScreenRenderingEnabled` instead of passing it around everywhere

r-ralph: Exactly! I'll address it :)

gpeal:
@r-ralph I went through the snapshots, marked some as duplicates but take a look at the latest ones. It looks like there is an issue with repeater opacity. Could you take a look?

r-ralph:
@gpeal There was a problem with creating the list `drawingContents`, so I reverted it. Thanks for confirmation.

gpeal:
@r-ralph Not sure why but your last commit didn't kick off snapshot tests on CI. Mind pushing a new empty commit?

r-ralph:
hmm,,,,, the task for uploading snapshot is skipped.
`Skipping snapshot deployment: was pull request.`

r-ralph:
oops, It is uploaded properly :)

gpeal:
@r-ralph it's an overloaded word 😉 
That is for maven snapshots
The visual snapshots run on a different CI server

lottie/src/main/java/com/airbnb/lottie/animation/content/ContentGroup.java
```diff
@@ -153,21 +160,48 @@ Matrix getTransformationMatrix() {
...
+        ((DrawingContent) content).draw(canvas, matrix, childAlpha);
+      }
+    }
+
+    if (isRenderingWithOffScreen) {
+      canvas.restore();
+    }
+  }
+
+  private boolean hasTwoOrMoreDrawableContent() {
```
gpeal: @r-ralph This function name isn't consistent with what it's doing.
However, it does seem smart to not do off screen rendering if there is only one drawing content. That's just not what this function is doing.

r-ralph: Sorry, my understanding might be wrong.
This function only checks the list `contents`, and I think it is consistent with the name. 
It is the caller's responsibility to decide whether to do off-screen drawing with this information, so I think it should not be changed to a name related to offscreen.

gpeal: @r-ralph Ah, I see how this function works. It's just a bit confusing the way it's written because it is called `hasTwoOrMoreDrawableContent` but has a variable `hasOneOrMoreDrawableContent` that gets set once and then returns true the next iteration with `DrawingContent`. Looks like it would work but it's confusing to read. What if you just had an int counter, increment it at the beginning of the instanceof block and then return right after if it's `>= 2`. I think that would be easier to read. What do you think?

r-ralph: I got it! I'll address it.

samiam2020:
@gpeal @r-ralph 
Can you check me on this but I think this solves the overlapping transparency issue. I just adjust the transparency on an alpha matte layer opposed to adjusting the transparency on the precomp. 

Example: 
1. INITIAL SET UP
LAYER 1 - PRE-COMP with opacity fade up 0%->100%
DESCRIPTION OF PRE-COMP contents is a character with multiple layered shape layers animating
RESULT on ANDROID: 
all the layers opacity fades up individually and you can see the layering.

2. NEW SET UP
LAYER 1 - SQUARE SHAPE LAYER with opacity fade up 0%->100% (being used as alpha matte)
LAYER 2 - PRE-COMP using LAYER 1 as alpha matte.
RESULT on ANDROID:
fades as expected not revealing layer overlaps.

PR #1372, opened 2019-08-29
Fix rounding-off error on progress calculation with time-remapped layer

r-ralph:
In #1368, progress calculation problem with time-remapping is fixed but it still have a possibility for rounding-off error.
I can't prepare a simple sample for reproduce, but I encount blinking issue with a company internal resource.


r-ralph:
In some test cases, it look wrong. I'll check.
https://happo.io/a/27/p/27/compare/master-android28/dad91db972f1658e3b18a70d9b75d9cf6289b55d-android28#prod-com.microblink.photomath-positive_rocket_animation

r-ralph:
I've found that the duration of compositon has offset in #766.
The time-remapped value is based on the original value(no-offset), so calculate with original value.

Issue #1380, opened 2019-09-02
Unused value


QiAnXinCodeSafe:
lottie-android/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java Lines 251 and 252
![image](https://user-images.githubusercontent.com/39950310/64108216-ddc21b00-cdae-11e9-9eb3-08f5d7baf65e.png)
Unused value

gpeal:
Fixed by #1385

PR #1385, opened 2019-09-09
resizing bitmap function as per issue #1380

jyotirmoy-paul:
using 'matrix approach' to resize the bitmap

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
@@ -250,7 +250,12 @@ public final class Utils {
     }
     float scaleWidth = ((float) width) / bitmap.getWidth();
     float scaleHeight = ((float) height) / bitmap.getHeight();
-    Bitmap resizedBitmap = Bitmap.createScaledBitmap(bitmap, width, height, true);
+
+    Matrix matrix = new Matrix();
+    matrix.postScale(scaleWidth, scaleHeight);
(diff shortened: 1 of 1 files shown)


lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
```diff
@@ -250,7 +250,12 @@ public static Bitmap resizeBitmapIfNeeded(Bitmap bitmap, int width, int height)
     }
     float scaleWidth = ((float) width) / bitmap.getWidth();
     float scaleHeight = ((float) height) / bitmap.getHeight();
-    Bitmap resizedBitmap = Bitmap.createScaledBitmap(bitmap, width, height, true);
+
+    Matrix matrix = new Matrix();
+    matrix.postScale(scaleWidth, scaleHeight);
+
+    Bitmap resizedBitmap = Bitmap.createBitmap(bitmap, 0, 0, width, height, matrix, false);
```
gpeal: @jyotirmoy-paul Can you just delete `scaleWidth` and `scaleHeight`? I think `createScaledBitmap` with the parameter values will be sufficient.

jyotirmoy-paul: @gpeal I have deleted ```scaleWidth``` and ```scaleHeight```. Now using ```createScaledBitmap``` to resize the bitmap directly. Thanks for suggesting.

PR #1397, opened 2019-09-28
Early return for GammaEvaluator.evaluate()

akihito104:
if `startInt` and `endInt` are same valule, the method always returns `startInt` (or `endInt`) value.


gpeal:
@akihito104 Can you rebase from master. The build was fixed and it'll show snapshot diffs.

akihito104:
was my operation wrong...? please tell me what to do.

PR #1406, opened 2019-10-09
Add sourceJar task.

boduan402:
Make it easy for user to build sourceJar file at local environment.


gpeal:
@boduan402 Why do you need this?

boduan402:
> @boduan402 Why do you need this?

For upgrade our Lottie version internally, I need to create sourceJar and jar locally. This cmd make this job easier.

PR #1418, opened 2019-11-02
Support ScaleType.FIT_XY in Lottie.

boduan402:
Support ScaleType.FIT_XY in LottieDrawable. 

Use the same way used in draw() method to do extra scale.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
@@ -8,11 +8,13 @@ import android.graphics.Canvas;
 import android.graphics.ColorFilter;
 import android.graphics.Matrix;
 import android.graphics.PixelFormat;
+import android.graphics.Rect;
 import android.graphics.Typeface;
 import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
 import android.os.Build;
 import android.view.View;
+import android.widget.ImageView;
 
 import androidx.annotation.FloatRange;
 import androidx.annotation.IntDef;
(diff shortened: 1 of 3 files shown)


gpeal:
This @boduan402 If you add tests here, they will show up in the snapshot diffs automatically
https://github.com/airbnb/lottie-android/blob/master/LottieSample/src/androidTest/java/com/airbnb/lottie/samples/LottieTest.kt#L292

I'll try to find some time to review this soon,

boduan402:
Thanks! No rush, I will add some test soon.

boduan402:
Hey Gabriel, can you suggest how to run screenshot test locally? 

The cmd suggested in comments seems out of date:

> Run these with: ./gradlew recordMode screenshotTests

boduan402:
Oh, I saw the 3 added ones here:
https://happo.io/a/27/p/27/compare/master-android28/ad53820ed121c566a4fa910b813caaebac69fa6e-android28?t=added

gpeal:
@boduan402 It requires API keys so you can't, unfortunately. You can keep pushing branches though.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -260,6 +265,10 @@ public void setApplyingOpacityToLayersEnabled(boolean isApplyingOpacityToLayersE
     this.isApplyingOpacityToLayersEnabled = isApplyingOpacityToLayersEnabled;
   }
 
+  public void setExtraScaleEnabled(boolean isEnabled) {
```
gpeal: @boduan402 Could you add some documentation around what this does since it is a public API? I'm actually not sure myself when end users would need to use this.

boduan402:
Sure, I am working on creating some tests to demonstrate why needs this api and will document its usage.

boduan402:
Gabriel, pls ignore the commits for testing the new screenshots. I will be cleaning up once figuring it out.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
boduan402: I have add doc in latest commit. Thanks!

Issue #1432, opened 2019-11-14
Crash in v3.2.0 if android:scaleType="fitCenter" is set


zboing:
If the user uses 3.2.0 dependency and uses the below code in the layout xml he will notice that the app crashes with below lines.
In version 3.1.0 this crash doesn't occur.

        <com.airbnb.lottie.LottieAnimationView
            android:id="@+id/soccer"
            android:layout_width="match_parent"
            android:layout_height="300dp"
            android:layout_centerInParent="true"
            android:layout_margin="20dp"
            android:scaleType="fitCenter"
            app:lottie_autoPlay="true"
            app:lottie_fileName="soccer-field.json"
            app:lottie_loop="true" />

```
     Caused by: android.view.InflateException: Binary XML file line #15: Error inflating class com.airbnb.lottie.LottieAnimationView
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.airbnb.lottie.LottieDrawable.setScaleType(android.widget.ImageView$ScaleType)' on a null object reference
        at com.airbnb.lottie.LottieAnimationView.setScaleType(LottieAnimationView.java:847)
 E/AndroidRuntime:     at android.widget.ImageView.<init>(ImageView.java:171)
        at android.widget.ImageView.<init>(ImageView.java:145)
        at androidx.appcompat.widget.AppCompatImageView.<init>(AppCompatImageView.java:72)
        at androidx.appcompat.widget.AppCompatImageView.<init>(AppCompatImageView.java:68)
        at com.airbnb.lottie.LottieAnimationView.<init>(LottieAnimationView.java:124)
        	... 32 more
```

boduan402:
I will send out a PR to fix this soon.

aymeekouakou:
Hi,
I have the same issue. Work perfectly in 2.8.0

boduan402:
PR is under review.

somesh-basicx:
Will there be a bug fix release right after this PR?

AndroidDeveloperLB:
Just a note:
This occurs on centerCrop scaleType too, as I wrote here (with sample) : 
https://github.com/airbnb/lottie-android/issues/1435

PattaFeuFeu:
This also occured for me with `centerInside` as the `scaleType`.

```xml
<com.airbnb.lottie.LottieAnimationView
    android:id="@+id/lav_animation"
    android:layout_width="0dp"
    android:layout_height="0dp"
[...]

PR #1433, opened 2019-11-15
Fix NPE when set ScaleType in xml.

boduan402:
Fix https://github.com/airbnb/lottie-android/issues/1432.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -204,6 +204,9 @@ import static com.airbnb.lottie.RenderMode.HARDWARE;
       setRenderMode(RenderMode.values()[renderModeOrdinal]);
     }
 
+    if (getScaleType() != null) {
+      lottieDrawable.setScaleType(getScaleType());
+    }
     ta.recycle();
 
     lottieDrawable.setSystemAnimationsAreEnabled(Utils.getAnimationScale(getContext()) != 0f);
@@ -844,7 +847,9 @@ import static com.airbnb.lottie.RenderMode.HARDWARE;
 
   @Override public void setScaleType(ScaleType scaleType) {
     super.setScaleType(scaleType);
-    lottieDrawable.setScaleType(scaleType);
+    if (lottieDrawable != null) {
+      lottieDrawable.setScaleType(scaleType);
+    }
   }
 
   @MainThread


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -844,7 +847,9 @@ public float getScale() {
 
   @Override public void setScaleType(ScaleType scaleType) {
     super.setScaleType(scaleType);
-    lottieDrawable.setScaleType(scaleType);
+    if (lottieDrawable != null) {
+      lottieDrawable.setScaleType(scaleType);
```
gpeal: @boduan402 Could this cause the scale type to never be set?

boduan402: It won't. 
- For case user config scaleType via XML, the lottieDrawable is null because it is called from super class's constructor. But the scaleType will be set into LottieDrawable at line 208 which is inside init() method.
- For case user set ScaleType through API call, the lottieDrawable is NonNull.

boduan402:
Hi Gabriel, do you think it's good to go?

droidluv:
The authors of lottie should seriously consider migrating to Kotlin to prevent these type of null safety related error's, I assume lottieDrawable variable is ever supposed to be null, its being globally initialised and suddenly it becomes null somewhere?

gpeal:
@droidluv Kotlin actually wouldn't have helped here. lottiDrawable is a final field. The issue is the order of instantiation between a constructor and fields which you can read more about [here](https://www.geeksforgeeks.org/order-execution-initialization-blocks-constructors-java/).

I'm about as big of a Kotlin fan as anybody out there and have been using it as my primary language for 4 years now. However, a major selling point for Lottie has been its lack of transitive dependencies. It used to have 0 and now it has 1, okio (which is tiny).

I've considered the Kotlin migration for a long time but have held off for a few reasons:
1) To avoid adding the kotlin stdlib as a dependency. However, as Kotlin adoption has skyrocketed and other libraries like okhttp have gone to Kotlin, I'd reconsider this.
2) A lot of work has been done to remove all autoboxing. While it's definitely possible to get to exactly the same level with Kotlin, it's easier to accidentally introduce autoboxing, especially when using nullables.
3) It would take many hours of work. I support Lottie on nights and weekends and this time would come directly from the work I do fixing bugs, adding features, and supporting the community.

Hope that clarifies things.

droidluv:
@gpeal thanks for the explanation, was quite humbling to realise that I went all these years as a programmer without understanding something seemingly very basic, the link you provided confused me a bit though because from their example, I already knew initialization went like 

1. Static Fields
2. Instance Fields
3. Constructor


What I didn't know was it was 
1. Static Fields
2. Instance Fields of Anscestors
3. Constructor of Anscestors
**4. Instance Fields of Current Class**
5. Constructor of Current Class

I had the notion that all instance fields get initialised before the constructors, which when I think about it makes less sense now, this article helped me a lot https://farenda.com/java/java-initialization-order/ Thanks Again, was truly an enlightening and humbling experience, and sorry for putting out the Kotlin demand, I am thankful that you make time to keep such an amazing library well maintained!

gpeal:
@droidluv no problem and there are always new things to learn!

PR #1446, opened 2019-11-30
solved TODO of Utils.getScale()

akihito104:
> figure out why the result needs to be divided by 2

the length of hypotenuse of isosceles right triangle that legs are sqrt(2) is 2. to get the length 1, legs are 1/sqrt(2).


akihito104:
if there is a problem with  calculation precision, please close this PR.

PR #1460, opened 2019-12-14
Improve primitive BaseKeyframeAnimation.getValue() performance

akihito104:
- introduced `cacheValue` for Int/Color/FloatKeyframeAnimation
- for process that calling directly `getIntValue()`/`getFloatValue()`, can be simply improved processing speed
- for process that calling `BaseKeyframeAnimation.getValue()`, to replace with `Utils.getIntValue()`/`getFloatValue()` makes  autoboxing/unboxing reduced

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/animation/keyframe/FloatKeyframeAnimation.java
+++ b/lottie/src/main/java/com/airbnb/lottie/animation/keyframe/FloatKeyframeAnimation.java
@@ -33,13 +33,20 @@ public class FloatKeyframeAnimation extends KeyframeAnimation<Float> {
       }
     }
 
-    return MiscUtils.lerp(keyframe.getStartValueFloat(), keyframe.getEndValueFloat(), keyframeProgress);
+    float value = MiscUtils.lerp(keyframe.getStartValueFloat(), keyframe.getEndValueFloat(), keyframeProgress);
+    cachedValue = value;
+    return value;
   }
 
   /**
    * Optimization to avoid autoboxing.
    */
+  private float cachedValue = 0f;
   public float getFloatValue() {
-    return getFloatValue(getCurrentKeyframe(), getInterpolatedCurrentKeyframeProgress());
+    float interpolatedProgress = getInterpolatedCurrentKeyframeProgress();
+    if (valueCallback == null && keyframesWrapper.isCachedValueEnabled(interpolatedProgress)) {
+      return cachedValue;
+    }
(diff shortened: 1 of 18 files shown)


lottie/src/main/java/com/airbnb/lottie/animation/keyframe/FloatKeyframeAnimation.java
```diff
@@ -33,13 +33,20 @@ float getFloatValue(Keyframe<Float> keyframe, float keyframeProgress) {
...
-    return MiscUtils.lerp(keyframe.getStartValueFloat(), keyframe.getEndValueFloat(), keyframeProgress);
+    float value = MiscUtils.lerp(keyframe.getStartValueFloat(), keyframe.getEndValueFloat(), keyframeProgress);
+    cachedValue = value;
+    return value;
   }
 
   /**
    * Optimization to avoid autoboxing.
    */
+  private float cachedValue = 0f;
```
gpeal: Can you move fields to the top for consistency with other classes?

lottie/src/main/java/com/airbnb/lottie/animation/keyframe/TransformKeyframeAnimation.java
```diff
@@ -172,12 +174,7 @@ public Matrix getMatrix() {
     }
 
     if (rotation != null) {
-      float rotation;
-      if (this.rotation instanceof ValueCallbackKeyframeAnimation) {
-          rotation = this.rotation.getValue();
-      } else {
-        rotation = ((FloatKeyframeAnimation) this.rotation).getFloatValue();
-      }
+      float rotation = getFloatValue(this.rotation);
```
gpeal: What about the `ValueCallbackKeyframeAnimation` case here?

akihito104: I think it can be called at following line of `Utils.getFloatValue()`.
https://github.com/airbnb/lottie-android/pull/1460/files/2f7643de191d29d1e45ba6cee478e1601e822d45#diff-cff0e0e4c6d992a385ff23491aecccc5R288

gpeal:
@akihito104 Do you have any benchmarks for the performance delta here?

akihito104:
I checked performance with some animations but the preferred effect was not obtained in almost cases. If just wants to suppress autoboxing, is it enough to call `Utils.getFloat/IntValue()`? if it was not needed, please close this PR.

akihito104:
my this code is no longer useful so i would close this pull request, sorry.

PR #1488, opened 2020-01-23
Upgraded Modules and MutablePair

Raks110:
**Upgraded Modules**
1. Modules have all been upgraded to the latest version.
2. A few modules used the '+' notation which can lead to unpredictable debug results as the latest version will always be used (which may be undesirable at times)

**Minor Change in MutablePair**
1. Simplified the equation in MutablePair.java:42


Raks110:
The checks have failed. Changes will be re-aimed at carefully.

Issue #1289, opened 2019-06-26
Add cache strategy feature


fleficher:
Hello there 👋 

It seems that it's not possible to control how animations are cached by Lottie.

I'm using multiple Lottie animations in my app. All of them are cached in memory and take a lot a memory involving OutOfMemory errors. Currently, we can't control which animation are cache in memory or disable this feature completely.

I've seen a `lottie_cacheStrategy` attribute in LottieAnimationView code but it doesn't seem to exist anymore 😢 . Why has it been removed?

Here's a sample of a memory dump showing the excessive amount of memory used by Lottie:

![58666818-b46b9c00-8334-11e9-8d96-00070946de70](https://user-images.githubusercontent.com/6464646/60214104-23211400-9865-11e9-84bd-19e2165b5dc5.png)

fleficher:
@gpeal why closing this issue? There's no provided solution at the moment :/

sunmTree:
any way to clear LottitCompositionCache?

kartik1225:
I am also looking for a solution for this.

axitasavani:
Is there find any solution for clear LottitCompositionCache?

sunmTree:
> Is there find any solution for clear LottitCompositionCache?

You can use Java Reflection to invoke "LottieCompositionCache.clear()" , but not sure influence

kartik1225:
@sunmTree  can you please give me a little bit more details on this?
I am less familiar with the Reflection API.

sunmTree:
> @sunmTree can you please give me a little bit more details on this?
> I am less familiar with the Reflection API.

Like this

```
try {
         Class<?> lottieCacheClass = Class.forName("com.airbnb.lottie.model.LottieCompositionCache");
            Method instance = lottieCacheClass.getDeclaredMethod("getInstance");
            Method clearMethod = lottieCacheClass.getMethod("clear", null);
            clearMethod.invoke(instance.invoke(null, null));
        } catch (ClassNotFoundException | NoSuchMethodException
                | IllegalAccessException | InvocationTargetException e) {
            e.printStackTrace();
        }
```

[User Java Reflection](https://www.oracle.com/technical-resources/articles/java/javareflection.html)

sunmTree:
> @sunmTree can you please give me a little bit more details on this?
> I am less familiar with the Reflection API.

Like this:

```
try {
            Class<?> lottieCacheClass = Class.forName("com.airbnb.lottie.model.LottieCompositionCache");
[...]

PR #1502, opened 2020-02-16
Two cache fixes

gpeal:
1. Add the ability to skip the network cache
2. Add the ability to clear all Lottie caches
Fixes https://github.com/airbnb/lottie-android/issues/1289


lottie/src/main/java/com/airbnb/lottie/network/NetworkCache.java
```diff
@@ -18,13 +18,24 @@
...
-    this.url = url;
+  }
+
+  public void clear() {
+    File parentDir = parentDir();
+    if (parentDir.exists()) {
+      File[] files = parentDir.listFiles();
+      if (files != null && files.length > 0) {
+        for (File file : parentDir.listFiles()) {
+          file.delete();
```
quochungnguyen97, line 34: I think the processing file should be handled on background thread to avoid ANR

gpeal: It is

quochungnguyen97: I think it isn't because the clear cache method of `LottieCompositionFactory` required to call in main thread to avoid multithread modification Collection of Map so the clear cache of NetworkCache bellow called in main thread too
```java
  public static void clearCache(Context context) {
    taskCache.clear();
    LottieCompositionCache.getInstance().clear();
    new NetworkCache(context).clear();
  }
```

PR #1532, opened 2020-03-13
WIP: blur layer effect

MightySeal:
Recently I found out that `Paint` supports mask filters, specifically [blur mask filter](https://developer.android.com/reference/android/graphics/BlurMaskFilter).

Several obvious drawbacks:
- After effects provides different options to handle edge pixels and also blur direction (vertical, horizontal, both) which is not possible to support with mask filter
- On some older devices will need to switch to software layer type
- Probably not possible to support several effects at once. Anyway, there are only 2 mask filters, but still.

Here is a proof of concept for fill and stroke content, also I've added blurred version of waving android so you can have a look, what do you think?


gpeal:
Interesting start! If you put the json file in the assets/tests folder, it will automatically get included in regression tests.

Test failures are not public because they require API keys but  [LottieLogo1.json](https://github.com/airbnb/lottie-android/blob/master/LottieSample/src/main/assets/Lottie%20Logo%201.json
) failed to parse.

gpeal:
@MightySeal Are you still planning on working on this?

PR #1557, opened 2020-04-29
(WIP) Support mirror in LottieDrawable.

boduan-coder:
Create an API to allow Lottie to mirror animation (text is not supported.)


boduan-coder:
We decided to go another approach https://github.com/airbnb/lottie-android/pull/1559

PR #1572, opened 2020-06-01
Enable xml preview in Android Studio.

pokercc:
image

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -317,7 +317,7 @@ import static com.airbnb.lottie.RenderMode.HARDWARE;
 
   @Override protected void onAttachedToWindow() {
     super.onAttachedToWindow();
-    if (!isInEditMode() && autoPlay || wasAnimatingWhenDetached) {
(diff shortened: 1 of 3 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -318,7 +317,7 @@ protected void onVisibilityChanged(@NonNull View changedView, int visibility) {
 
   @Override protected void onAttachedToWindow() {
     super.onAttachedToWindow();
-    if (autoPlay || wasAnimatingWhenDetached) {
+    if (!isInEditMode() && autoPlay || wasAnimatingWhenDetached) {
```
gpeal, line 320: I don't think this is right. You want parens around the two ORs

pokercc: Thanks a lot.  I had fix this. Please review again.

IgorCordasEndava:
This feature is awesome, thank you for this PR. 
We have been in need for this for like 2 years, most of our lottie views had tools:src with first frame as png and now we don't have to do this anymore. Developing this further (set preview frame, maybe even animate) would definitely be even more useful.

PR #1575, opened 2020-06-08
Fix blinking black in multithreading mode

tttzof351:
I needed to overlay the animation on the video in the background thread and I encountered a black blinking due to the presence of a shared static state in Utils


tttzof351:
@gpeal 
Sorry for a long answer - I wrote comment

Issue #1310, opened 2019-07-10
Using Text Delegate and scaletype then the textfield wont position correctly


Jona85:
When I use TextDelegate to change the text in my animation and I use a scaletype for example "Center" then the textfield doesnt respect the scaling & positioning of the LottieAnimationView. 
I've tried to workaround the issue with a nulllayer in After Effects, but this didnt resolve the issue.

When I enable the Glyphs option in BodyMovin then the text is positioned as expected, but then I wont be able to change it with TextDelegate. 

[lottie_text_delegate_test.zip](https://github.com/airbnb/lottie-android/files/3377976/lottie_text_delegate_test.zip)

Jona85:
As for now I've "fixed" it with a workaround.
In AfterEffects I added an invisible textlayer containing all the possible characters. 
Enabled the Glyphs option in Bodymovin.
In our Android project instead of using the TextDelegate I read the animation.json from the assets and use replace() to update the text. 
Then I use the updated json string to load into my LottieAnimationView.

himanshusaluja:
Facing same issue text alignment fails on different screen density looks like lottie text elements is not scaling as per screen dimensions.Is their any solution for this

volodia-chornenkyy:
This fixed issue for me if view or parent has a fixed size.
`app:lottie_scale="2"`

sweetbot:
I had the same issue. This issue only happened when using TextDelegate on Android, not on iOS. 
Solution: I added 2 lines in xml:
android:adjustViewBounds="true"
android:scaleType="centerCrop"
So my lottie animation view looks like this:
```
    <com.airbnb.lottie.LottieAnimationView
        android:id="@+id/animation_view"
        android:layout_width="match_parent"
        android:layout_height="match_parent"
        android:adjustViewBounds="true"
        android:scaleType="centerCrop" />
```
This will solve the wrong text positioning problem perfectly when when using TextDelegate on Android.
Hope this helps.

shirley737:
Hello @sweetbot , can you show me an example of how to use textDelegate method? I haven't found any documentation yet...Thank you!

PR #1583, opened 2020-06-20
fixed: Using Text Delegate and scaletype then the textfield wont posi…

pikaMouse:
I had the same issue as #1310. When I use TextDelegate to change the text in my animation and I use a scaletype for example "Center" then the textfield doesn't respect the scaling & positioning of the LottieAnimationView.At present，I've fixed it with using "concat" rather than "setMatrix" before draw something in the canvas. The difference between them as follow：
* setMatrix:Completely replace the current matrix with the specified matrix.
* concat:Preconcat the current matrix with the specified matrix


mgod:
This fixed some text layout issues I was having as well. If we get this or a similar PR passing on the master branch, is it likely to get merged?

pikaMouse:
> This fixed some text layout issues I was having as well. If we get this or a similar PR passing on the master branch, is it likely to get merged?

I try it

mgod:
@pikaMouse I hope you don't mind, I included this commit along with some other dynamic text layout fixes in PR #1635

pikaMouse:
> @pikaMouse I hope you don't mind, I included this commit along with some other dynamic text layout fixes in PR #1635

Of course not

PR #1635, opened 2020-09-30
Fix scaling and tracking for dynamic text

### LINKED ISSUE #1183: TextLayer uses incorrect textLineWidth to draw
Hi, I am trying to debug the TextLayer.java, I find the `textLineWidth` which used to do text justification may be incorrect.  The `textLineWidth` never consider the `documentData.tracking `value.
image

### LINKED ISSUE #1310: Using Text Delegate and scaletype then the textfield wont position correctly

mgod:
Close #1183. Close #1310.

This has a couple of related bugfixes.

1. Fix the scale/positioning of the text relative to the parent (from PR #1583, thanks @pikaMouse)
1. Include text tracking values in the line width used for text alignment
1. Scale the text tracking values based on the font size rather than the parent size (parent size adjustment is already handled when we're calculating tracking values)
1. Include all bytes for multibyte characters when calculating canvas offsets for characters.

From #1183, here's the updated test animation result with the text correctly centered between the lines:

Screen Shot 2020-09-30 at 11 02 54 AM

And here's the before and after for the same sample with some 🦁 emojis added in to show multibyte character measurement:

Before:
Screen Shot 2020-09-30 at 11 10 47 AM

After:
Screen Shot 2020-09-30 at 11 11 43 AM

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/model/layer/TextLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/model/layer/TextLayer.java
@@ -116,7 +116,7 @@ public class TextLayer extends BaseLayer {
   void drawLayer(Canvas canvas, Matrix parentMatrix, int parentAlpha) {
     canvas.save();
     if (!lottieDrawable.useTextGlyphs()) {
-      canvas.setMatrix(parentMatrix);
+      canvas.concat(parentMatrix);
     }
     DocumentData documentData = textAnimation.getValue();
     Font font = composition.getFonts().get(documentData.fontName);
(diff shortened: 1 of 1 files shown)


gpeal (approved):
The diffs look great. Thanks!

PR #1636, opened 2020-10-01
Clamp gradient color opacity

AlexKGwyn:
Fixes an issue where color positions that came before any opacity positions in a gradient color array were interpolated to negative values, throwing off the conversion into a color.

An example of a gradient color array that presented this issue:
"k": [0, 0.969, 0.514, 0.745, 0.116, 0.984, 0.537, 0.373, 1, 1, 0.559, 0, 0.607, 0, 0.776, 0.5, 0.945, 1]

**Colors**
0, 0.969, 0.514, 0.745,
0.116, 0.984, 0.537, 0.373,
1, 1, 0.559, 0,
**Opacity**
0.607, 0, 
0.776, 0.5, 
0.945, 1

Since the first color came before the first opacity position, the progress was calculated as about -3.59 causing the following linear interpolation to give out a large negative value which broke things when it was bit shifted into the alpha position for the final color.

Here are some screenshots:

**Original issue**
Screenshot_20201001-011043

**Resolved**
Screenshot_20201001-010909

**Sample Json** (as a txt because github doesn't allow json)
[testjson.txt](https://github.com/airbnb/lottie-android/files/5310542/testjson.txt)


lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
```diff
@@ -139,7 +139,7 @@ private int getOpacityAtPosition(double position, double[] positions, double[] o
       double lastPosition = positions[i - 1];
       double thisPosition = positions[i];
       if (positions[i] >= position) {
-        double progress = (position - lastPosition) / (thisPosition - lastPosition);
+        double progress = MiscUtils.clamp((position - lastPosition) / (thisPosition - lastPosition), 0, 1);
```
AlexKGwyn, line 142: I assume it would be fine to change the precision of the progress and use an existing clamp method. I didn't want to make that change without some more knowledge about the choices here though.

gpeal:
@AlexKGwyn This looks great. I verified that the diffs in the link above are, in fact, more accurate that the originals. Could you also add your json file to [this folder](https://github.com/airbnb/lottie-android/tree/master/LottieSample/src/main/assets/Tests). That will automatically include it in the report above.

gpeal (approved):
Thanks!

PR #1660, opened 2020-10-27
add support for dotLottie 

kudanai:
Fix parsing .lottie files
the dotLottie format  (dotlottie.io) is a zip file with animation.json bundled along with image resources, and manifest

**Changes**
* update zipStreamSync() to handle dotLottie use case
  * ignore manifest.json otherwise lottie will try to use it

* update network loader to check for .lottie and treat it as a .zip

* add support for .lottie and in rawRes
  * use magic header to determine files and use zipStreamSync() instead of fromJsonSync()
  * this has the bonus side effect of allowing .zip files in rawRes as well (which doesn't appear to have been supported before)

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -487,7 +487,7 @@ public class LottieCompositionFactory {
     try {
       BufferedSource peek = inputSource.peek();
       for (byte b: MAGIC) {
-        if(peek.readByte()!=b)
+        if(peek.readByte() != b)
           return false;
       }
       peek.close();
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -49,6 +49,13 @@ public class LottieCompositionFactory {
    */
   private static final Map<String, LottieTask<LottieComposition>> taskCache = new HashMap<>();
 
+  /**
+   * reference magic bytes for zip compressed files.
+   * useful to determine if an InputStream is a zip file or not
+   */
+  private static final int[] MAGIC = new int[] { 0x50, 0x4b, 0x03, 0x04 };
+
+
   private LottieCompositionFactory() {
   }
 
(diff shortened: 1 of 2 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -465,6 +479,26 @@ private static boolean isNightMode(Context context) {
...
   }
 
+  /**
+   * Check if a given InputStream points to a .zip compressed file
+   */
+  private static Boolean isZipCompressed(BufferedSource inputSource) {
+
+    try {
+      BufferedSource peek = inputSource.peek();
+      for(byte b: MAGIC) {
```
gpeal, line 489:
```suggestion
      for (byte b: MAGIC) {
```

lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -465,6 +479,26 @@ private static boolean isNightMode(Context context) {
...
+        if(peek.readByte()!=b)
```
gpeal, line 490:
```suggestion
        if(peek.readByte() != b)
```

lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -253,7 +261,11 @@ public static void clearCache(Context context) {
   @WorkerThread
   public static LottieResult<LottieComposition> fromRawResSync(Context context, @RawRes int rawRes, @Nullable String cacheKey) {
     try {
-      return fromJsonInputStreamSync(context.getResources().openRawResource(rawRes), cacheKey);
+      BufferedSource source = Okio.buffer(source(context.getResources().openRawResource(rawRes)));
+      if(isZipCompressed(source)) {
```
gpeal, line 265:
```suggestion
      if (isZipCompressed(source)) {
```

lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -465,6 +479,26 @@ private static boolean isNightMode(Context context) {
...
+    try {
+      BufferedSource peek = inputSource.peek();
+      for(byte b: MAGIC) {
+        if(peek.readByte()!=b)
+          return false;
+      }
+      peek.close();
+      return true;
+    } catch (Exception e) {
+      e.printStackTrace();
```
gpeal, line 496:
```suggestion
      L.error(e, "Failed to check zip file header")
```

gpeal: `e.printStackTrace()` should never be used on Android

gpeal (commented):
Looks great! Just a few minor nits

gpeal (approved):

PR #1662, opened 2020-10-31
Add Animator.AnimatorPauseListener support

### LINKED ISSUE #1644: addAnimatorPauseListener() support

Using lottie for Android 3.4.4, I'm trying to play a sound in sync with an animation. But I didn't find a reliable way to detect when the animation pause/resume, as the `LottieAnimationView` does not provide callback on these actions.

Android provide an `Animator.AnimatorPauseListener` interface alongside `Animator.AnimatorListener`.

`LottieAnimationView` API already supports `Animator.AnimatorListener`, but it does not supports `Animator.AnimatorPauseListener`

Supporting this kind of listener would be a natural extension of the `LottieAnimationView` API.

`LottieAnimationView.addAnimatorPauseListener()`
`LottieAnimationView.removeAnimatorPauseListener()`

Actually, I'm subclassing LottieAnimationView to generate and dispatch pause/resume events when pauseAnimation()/resumeAnimation() is invoked, but it does not allow me to reuse the `Animator.AnimatorPauseListener` interface as I don't have access to the `ValueAnimator` that should be passed to the callback, and it would be better if implemented directly on the underlying drawable.

heitorcolangelo:
Resolves #1644 by adding methods to LottieAnimationView and LottieDrawable

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
@@ -743,21 +743,11 @@ public class LottieDrawable extends Drawable implements Drawable.Callback, Anima
     animator.removeAllListeners();
   }
 
-  public void addAnimatorPauseListener(Animator.AnimatorPauseListener listener) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-      animator.addPauseListener(listener);
-    } else {
-      Logger.warning("Add AnimatorPauseListeners are not supported pre-Kit Kat.");
-    }
-  }
+  @RequiresApi(api = Build.VERSION_CODES.KITKAT)
+  public void addAnimatorPauseListener(Animator.AnimatorPauseListener listener) { animator.addPauseListener(listener); }
 
-  public void removeAnimatorPauseListener(Animator.AnimatorPauseListener listener) {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-      animator.removePauseListener(listener);
-    } else {
-      Logger.warning("Remove AnimatorPauseListeners are not supported pre-Kit Kat.");
-    }
-  }
+  @RequiresApi(api = Build.VERSION_CODES.KITKAT)
+  public void removeAnimatorPauseListener(Animator.AnimatorPauseListener listener) { animator.removePauseListener(listener); }
 
   /**
    * Sets the progress to the specified frame.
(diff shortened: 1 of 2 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -742,6 +743,22 @@ public void removeAllAnimatorListeners() {
     animator.removeAllListeners();
   }
 
+  public void addAnimatorPauseListener(Animator.AnimatorPauseListener listener) {
```
gpeal, line 746: I think it's safer to add `@RequiresApi(api = Build.VERSION_CODES.KITKAT)` to the public API here. Most people don't pay attention to the logs.

gpeal (commented):
You'll also have to add the same corresponding methods in `LottieAnimationView` that just call through to `LottieDrawable` You can follow the pattern that lots of other methods there do.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
heitorcolangelo: I agree, should I remove the `if`statement then?

heitorcolangelo:
@gpeal I don't think I got what you mean by "add the same corresponding methods in `LottieAnimationView`", I did add two methods to `LottieAnimationView`. Can you please be more specific about what I'm missing?

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
gpeal: @heitorcolangelo Oh you're right. I hadn't had my coffee yet!

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -742,6 +743,22 @@ public void removeAllAnimatorListeners() {
...
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
gpeal, line 747: @heitorcolangelo We should add the same `@RequiresApi` here too. This is also a public API.

heitorcolangelo: @gpeal
 And do you think we can remove this if statement? I believe it is not necessary as we added the annotation.
The final version would be just
```kotlin
@RequiresApi(api = Build.VERSION_CODES.KITKAT)
public void addAnimatorPauseListener(Animator.AnimatorPauseListener listener) {
  animator.addPauseListener(listener);
}

@RequiresApi(api = Build.VERSION_CODES.KITKAT)
public void removeAnimatorPauseListener(Animator.AnimatorPauseListener listener) {
  animator.removePauseListener(listener);
}
```

gpeal (approved):

PR #1679, opened 2020-11-12
Add dynamic text style property

Luosoha:


lottie/src/main/java/com/airbnb/lottie/model/Font.java
```diff
@@ -9,7 +9,7 @@
 
   private final String family;
   private final String name;
-  private final String style;
+  private String style;
```
gpeal, line 12: This mutates the Font file on the composition which could be used across many layers or even reused across multiple animation views.
The specific layer with the callback will need to create its own immutable font to use if it differs from the one here.

lottie/src/main/java/com/airbnb/lottie/model/layer/TextLayer.java
```diff
@@ -174,6 +176,13 @@ private void drawTextGlyphs(
     } else {
       textSize = documentData.size;
     }
+    String textStyle = null;
```
gpeal, line 179: You only included support for glyphs. Can you include the code path that draws text with a font as well?

lottie/src/main/java/com/airbnb/lottie/model/layer/TextLayer.java
```diff
@@ -174,6 +176,13 @@ private void drawTextGlyphs(
...
+    if (textStyleCallbackAnimation != null) {
+      textStyle = textStyleCallbackAnimation.getValue();
```
gpeal, line 181: Can you update this so the callback is only called once per frame?

gpeal (commented):
Overall, I think this approach is more scaleable than your approach in https://github.com/airbnb/lottie-android/pull/1678 because it is designed to target specific elements rather than being set at the top level.

Plus, if we were to land both, the priority between the top level setter and dynamic property would be ambiguous.

In addition, please add a test in `LottieTest`. It will be run if you push a new commit (takes ~20-30 minutes) and will post a report with snapshots of the results like [this](https://github.com/airbnb/lottie-android/pull/1658#issuecomment-716189603)

lottie/src/main/java/com/airbnb/lottie/model/Font.java
Luosoha: @gpeal Does it really need to be immutable? Cuz I want to change it dynamically based on the user interests, and as they might change the style multiple times I need to change the font style variable as well.

lottie/src/main/java/com/airbnb/lottie/model/layer/TextLayer.java
Luosoha: @gpeal I did, please check line 269

Luosoha:
@gpeal Also can you help me with this feature? Sadly I can't seem to fulfill your requests :(

lottie/src/main/java/com/airbnb/lottie/model/Font.java
gpeal: @Luosoha the problem is that this font object will be reused by any other layers that use the same font so even if you set the KeyPath on a specific text layer, it will affect other layers with the same font.

Luosoha: @gpeal  Aight I understand it now, but can you finish this PR for me? Sadly I'm not capable of doing the works you are asking me for.

Luosoha:
@gpeal Hi, any news for supporting these?

gpeal:
Unfortunately, this isn't a priority for me right now. If somebody wants to take this across the finish line, they can cherry pick these changes and continue it.

PR #1683, opened 2020-11-18
[Compose] Removed unnecessary dependencies

TonnyL:
`material` and `material-icons-android` these dependencies contain a giant number of icons which are not used in `lottie-compose` module, so I just removed them to reduce apk file size.


gpeal:
Ah, they were removed in https://github.com/airbnb/lottie-android/commit/08691b27335a345dc1437ae5ff9b4b2422758dfe. Good catch though!

PR #1706, opened 2020-12-25
Add an API to always render images at the original size

gpeal:
This allows apps that set custom images to provide images that are smaller than the original image to save memory.

Fixes https://github.com/airbnb/lottie-android/issues/1968


soongao:
the merge request was interrupted,

DirtyWhiteBoy34 (approved):

Issue #1734, opened 2021-02-01
Animation doesn't render correctly using Lottie-Compose


jossiwolf:
**Describe the bug**
With Lottie-Compose, the animation doesn't render correctly, some parts are missing. The same animation works with `LottieAnimationView`. Am I misusing `LottieAnimation`?

**What version of Lottie did you test this on?**
Lottie-Compose Alpha 05, 06-SNAPSHOT
Jetpack Compose Alpha 10, 11

I didn't test this with the latest snapshot yet because I didn't find the docs for that :D Will test that as soon as I figure out how to include the snapshot :) I don't expect it to change though

**What version of Android did you test this on?**
Pixel 4XL, Android 11

**Steps To Reproduce**

1. Use the attached `lightbulb.json` file in the zip (this is from https://lottiefiles.com/7955-lightbulb)
[7955-lightbulb.zip](https://github.com/airbnb/lottie-android/files/5903317/7955-lightbulb.zip)
2. Use `LottieAnimation` to display the animation

```kotlin
val animationSpec = LottieAnimationSpec.RawRes(R.raw.lightbulb)
val animationState = rememberLottieAnimationState(
    autoPlay = true,
    repeatCount = Int.MAX_VALUE
)
LottieAnimation(
    spec = animationSpec,
    animationState = animationState
)
```

Wrapping `LottieAnimationView` using `AndroidView` works as it should.

```kotlin
@Composable
fun AndroidLottieAnimationView(
    modifier: Modifier = Modifier,
    spec: LottieAnimationSpec.RawRes,
    animationState: LottieAnimationState
) {
    AndroidView(
        modifier = modifier,
        viewBlock = { context -> LottieAnimationView(context) },
        update = { view ->
            view.apply {
                setAnimation(spec.resId)
                repeatCount = animationState.repeatCount
                frame = animationState.frame
                if (animationState.isPlaying) playAnimation()
            }
        }
    )
}
```

**Screenshots/Videos**

|`LottieAnimationView`|`LottieAnimation`|
|---|---|
|![](https://user-images.githubusercontent.com/10628007/106445182-fd1b1700-647e-11eb-82db-a3730ddc282d.gif)|![](https://user-images.githubusercontent.com/10628007/106445006-ca711e80-647e-11eb-8cda-a8a7a30ea677.gif)|

PR #1744, opened 2021-02-14
[Compose] Add merge paths flag

jossiwolf:
In relation to #1734

`LottieAnimationView` had merge paths disabled by default, but `LottieAnimation` enables them by default. As long as merge paths contain known bugs, I think it makes sense not to change that behavior - what do you think?
I added a flag for it to the state. If you want, we can add it as a constructor param too, but I feel like that's not needed as it's not that commonly used.

Cheers

### DIFF
--- a/lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
+++ b/lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
@@ -120,12 +120,10 @@ fun LottieAnimation(
             .then(modifier)
     ) {
         drawIntoCanvas { canvas ->
-            drawable.apply {
-                progress = state.progress
-                setOutlineMasksAndMattes(state.outlineMasksAndMattes)
-                isApplyingOpacityToLayersEnabled = state.applyOpacityToLayers
-                enableMergePathsForKitKatAndAbove(state.enableMergePaths)
-            }
+            drawable.progress = state.progress
+            drawable.setOutlineMasksAndMattes(state.outlineMasksAndMattes)
+            drawable.isApplyingOpacityToLayersEnabled = state.applyOpacityToLayers
+            drawable.enableMergePathsForKitKatAndAbove(state.enableMergePaths)
             withTransform({
                 scale(size.width / composition.bounds.width().toFloat(), size.height / composition.bounds.height().toFloat(), Offset.Zero)
             }) {
@@ -140,4 +138,4 @@ private fun Modifier.maintainAspectRatio(composition: LottieComposition?): Modif
     return this.then(aspectRatio(composition.bounds.width() / composition.bounds.height().toFloat()))
 }
 
-private fun lerp(a: Float, b: Float, @FloatRange(from = 0.0, to = 1.0) percentage: Float) =  a + percentage * (b - a)
+private fun lerp(a: Float, b: Float, @FloatRange(from = 0.0, to = 1.0) percentage: Float) = a + percentage * (b - a)
--- a/lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
+++ b/lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
@@ -82,7 +82,7 @@ fun LottieAnimation(
 ) {
     val drawable = remember {
         LottieDrawable().apply {
-            enableMergePathsForKitKatAndAbove(true)
+            enableMergePathsForKitKatAndAbove(state.enableMergePaths)
         }
     }
 
@@ -118,7 +118,6 @@ fun LottieAnimation(
     }
 
     if (composition == null || composition.duration == 0f) return
-    SideEffect {}
 
     Canvas(
         modifier = Modifier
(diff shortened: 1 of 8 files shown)


lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimationState.kt
```diff
@@ -6,25 +6,49 @@ import androidx.compose.runtime.mutableStateOf
...
 }
 
 /**
+ * State of the [LottieAnimation] composable
+ *
+ * @param autoPlay Whether the animation will auto play
+ * @param repeatCount How many times the animation will be played. Use [Int.MAX_VALUE] for
+ * infinite repetitions.
+ * @param initialProgress The progress the animation starts with
+ * @param enableMergePaths Whether merge paths are enabled
```
gpeal, line 43: You can make these docs `Initial value for [propName]` so they are directly linked and the full docs can be there.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
```diff
@@ -82,7 +82,7 @@ fun LottieAnimation(
 ) {
     val drawable = remember {
         LottieDrawable().apply {
-            enableMergePathsForKitKatAndAbove(true)
+            enableMergePathsForKitKatAndAbove(state.enableMergePaths)
```
gpeal, line 85: This shouldn't be set here because it will only capture the initial state. I think it should be something like
```
SideEffect(state.enableMergePaths) {
    drawable.enableMergePathsForKitKatAndAbove = state.enableMergePaths
}
````

Comment (jossiwolf) [reply lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt]: Hah, totally missed that we're remembering this here.. thanks! I ended up pushing the read down to the `draw` block which lets us skip the recomposition.

**Comment by jossiwolf** [comment]
@gpeal I added a toggle for merge paths to the Compose sample. Let me know if you want to have that as part of this PR.


https://user-images.githubusercontent.com/10628007/108623448-720abc80-743f-11eb-8525-968b61439c6b.mp4

File: lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt [hunk 3]
```diff
@@ -118,17 +113,19 @@ fun LottieAnimation(
...
     Canvas(
         modifier = Modifier
             .maintainAspectRatio(composition)
             .then(modifier)
     ) {
         drawIntoCanvas { canvas ->
-            drawable.progress = state.progress
-            drawable.setOutlineMasksAndMattes(state.outlineMasksAndMattes)
-            drawable.isApplyingOpacityToLayersEnabled = state.applyOpacityToLayers
+            drawable.apply {
```
gpeal, line 123: nit: I don't think the drawable.apply makes things easier to read here. It just adds 2 lines of code an some cognitive overhead to figure out what the `this` is.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimationState.kt
```diff
@@ -6,33 +6,64 @@ import androidx.compose.runtime.mutableStateOf
...
 import androidx.compose.runtime.setValue
 
+/**
+ * Create a [LottieAnimationState] and remember it
+ *
+ * @param autoPlay Whether the animation will auto play
+ * @param repeatCount How many times the animation will be played. Use [Int.MAX_VALUE] for
+ * infinite repetitions.
+ * @param initialProgress The progress the animation starts with
+ * @param enableMergePaths Whether merge paths are enabled
```
gpeal, line 16: Can you update the docs here to match below?

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimationState.kt
```diff
@@ -6,33 +6,64 @@ import androidx.compose.runtime.mutableStateOf
...
+    enableMergePaths: Boolean = true
 ) {
     var progress by mutableStateOf(initialProgress)
 
     // TODO: make this public
     private var _frame = mutableStateOf(0)
     val frame: Int by _frame
 
+    /**
+     * Whether the animation is currently playing
```
gpeal, line 59:
```suggestion
     * Whether the animation is currently playing.
```

gpeal (commented):
Pretty close! Just a couple small comments.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
jossiwolf: Hmm I just compared and think it's easier to read that way - compare it to

```kotlin
drawable.progress = state.progress
drawable.setOutlineMasksAndMattes(state.outlineMasksAndMattes)
drawable.isApplyingOpacityToLayersEnabled = state.applyOpacityToLayers
drawable.enableMergePathsForKitKatAndAbove(state.enableMergePaths)
```

I'm expecting some other things will be added here too over time, so it makes sense to scope it to the `apply` IMO. Unless you have strong concerns over this I'll keep it :)

gpeal: @jossiwolf although there is no "right" here, I do prefer avoiding `apply` unless it is a very large block. The `drawable` prefix is roughly analogous to the extra indentation from the apply block but it makes it immediately clear what `this` is. Without it, you could silently mix and match calls this this@LottieDrawable and this@LottieAnimation which could lead to ambiguous situations.

jossiwolf: Yep, let's go with what's more agnostic to the codebase then! I'll push a commit tomorrow:)

gpeal (approved):
Thanks!

jossiwolf:
Awesome, thanks so much!

PR #1827, opened 2021-06-21
[Compose] Breaking Change: Major Compose API Refactor V2

gpeal:
This is a second take at https://github.com/airbnb/lottie-android/pull/1793 that much more closely resembles the [Animatable](https://developer.android.com/jetpack/compose/animation#animatable) APIs.

This PR is a major update to the Lottie Compose APIs.
Notable changes:
* `LottieAnimation` composable now just takes a composition, progress as a Float, and base properties such as ImageAssetDelegate, mergePaths, etc.
* Animating the composition can be done either with a manual animation, gesture, etc or via the new `animateLottieComposition()` function or `LottieAnimatable` class.
* LottieAnimation contains overloads that make it easier to just pass in a LottieCompositionSpec or animation parameters and under the hood, it will wrap `lottieComposition()` and `animateLottieComposition()`.

The next set of APIs after this PR will be dynamic properties.

## Manual Progress
```kotlin
val composition by rememberLottieComposition(LottieCompositionSpec.RawRes(R.raw.heart))
LottieAnimation(
    composition,
    progress = 0.5f,
)
```

## Animating Progress
```kotlin
// Driving an animation has been split out into animateLottieComposition(...)
// custom animation using Lottie's animator
var progress by animateLottieCompositionAsState(
    iterations = LottieConstants.IterateForever,
)
LottieAnimation(
    LottieCompositionSpec.RawRes(R.raw.heart),
    progress = progress,
)
```
```kotlin
// This will play one time whenever isPlaying becomes true
val isPlaying by remember { mutableStateOf(false) }
val composition by rememberLottieComposition(LottieCompositionSpec.RawRes(R.raw.heart))
LottieAnimation(
    composition,
    isPlaying = isPlaying,
)
```

## Using LottieAnimatable
```kotlin
val anim = rememberLottieAnimatable()
val composition by rememberLottieComposition(LottieCompositionSpec.RawRes(R.raw.heart))
LaunchedEffect(composition) {
    anim.animate(
        composition,
        iterations = LottieConstants.IterateForever,
    )
}
LottieAnimation(anim.composition, anim.progress)
```

```kotlin
val anim = rememberLottieAnimatable()
val composition by rememberLottieComposition(LottieCompositionSpec.RawRes(R.raw.heart))
var sliderGestureProgress: Float? by remember { mutableStateOf(null) }
LaunchedEffect(composition, sliderGestureProgress) {
    when (val p = sliderGestureProgress) {
        null -> anim.animate(
            composition,
            iterations = LottieConstants.IterateForever,
            initialProgress = anim.progress,
            continueFromPreviousAnimate = false,
        )
        else -> anim.snapTo(progress = p)
    }
}
Box {
    LottieAnimation(composition, anim.progress)
    Slider(
        value = sliderGestureProgress ?: anim.progress,
        onValueChange = { sliderGestureProgress = it },
        onValueChangeFinished = { sliderGestureProgress = null },
        modifier = Modifier
            .align(Alignment.BottomCenter)
            .padding(bottom = 8.dp)
    )
}
```

## Retries
There is now a simple way to retry animations that fail to load.
```kotlin
var backoffMs = 2000L
val composition by lottieComposition(
    LottieCompositionSpec.Url("https://...")
    onRetry = { retryCount, _ ->
        // This will do 6 exponential backoff retries.
        delay(backoffMs)
        backoffMs *= 2
        retryCount <= 5
    },
)
```

## Transitions
This also includes the ability to create transitions where different states map to different compositions or clip specs within an animation:
```kotlin
val composition by rememberLottieComposition(LottieCompositionSpec.RawRes(R.raw.bar))
val animatable = rememberLottieAnimatable()
val state by rememberUpdatedState(section)

LaunchedEffect(composition, animatable) {
    composition ?: return@LaunchedEffect
    snapshotFlow { state }.collectLatest { s ->
        val clipSpec = when (s) {
            TransitionSection.Intro -> LottieClipSpec.Progress(0f, 0.301f)
            TransitionSection.LoopMiddle -> LottieClipSpec.Progress(0.301f, 2f / 3f)
            TransitionSection.Outro -> LottieClipSpec.Progress(2f / 3f, 1f)
        }
        do {
            animatable.animate(
                composition,
                clipSpec = clipSpec,
                cancellationBehavior = LottieCancellationBehavior.OnIterationFinish,
            )
        } while (s == TransitionSection.LoopMiddle)
    }
}
LottieAnimation(composition, animatable.progress)
```


kenyee (approved):
FWIW, I like the new API better as well...feels more consistent w/ Compose APIs.
Thanks for making the changes.

sample-compose/src/main/java/com/airbnb/lottie/sample/compose/examples/AnimatableExamplesPage.kt
```diff
@@ -0,0 +1,110 @@
...
+private fun Example1() {
+    val anim = remember { LottieAnimatable() }
+    val composition by lottieComposition(LottieCompositionSpec.RawRes(R.raw.heart))
+    LaunchedEffect(composition) {
+        anim.animate(
+            composition,
+            iterations = LottieConstants.IterateForever,
+        )
+    }
+    LottieAnimation(composition, anim.progress)
```
doris4lt, line 47: This seems a little redundant and potentially easy to forget.  If a user has already created a lottie composition, and a LottieAnimatable, can the wire-up be done in either of those two steps instead of requiring an extra step? Animatable seems to have the info about both the composition, and progress.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
```diff
@@ -10,142 +11,226 @@ import androidx.compose.ui.graphics.drawscope.drawIntoCanvas
...
+/**
+ * This is like [LottieAnimation] except that it takes a [LottieCompositionSpec] instead of the
+ * [LottieComposition] itself. This will be more convenient for most use cases unless you
+ * need to explicitly check for failure states while loading the composition, which is done
+ * via [lottieComposition].
+ *
+ * @see LottieAnimation
+ */
+@Composable
+fun LottieAnimation(
```
doris4lt, line 127: This composable could use a different name, since it's entirely driven by a progress instead of owning the animation.

lottie-compose/src/main/java/com/airbnb/lottie/compose/animateLottieComposition.kt
```diff
@@ -0,0 +1,72 @@
...
+ *                      the progress and iteration gets reset.
+ * @param clipSpec A [LottieClipSpec] that specifies the bound the animation playback
+ *                 should be clipped to.
+ * @param speed The speed the animation should play at. Numbers larger than one will speed it up.
+ *              Numbers between 0 and 1 will slow it down. Numbers less than 0 will play it backwards.
+ * @param iterations The number of times the animation should repeat before stopping. It must be
+ *                    a positive number. [LottieConstants.IterateForever] can be used to repeat forever.
+ */
+@Composable
+fun animateLottieComposition(
```
doris4lt, line 34: What's the use case for this, or is it more of an impl detail? :-)

doris4lt (commented):
V2 is a lot more clear in terms of separation between the state-based use cases and imperative use cases. After reading the API offerings, I have a pretty good idea which APIs to reach for for a particular use case.  👍 

I would try reducing the amount of overloads when possible. You could consider changing the names of some of the overloads to something more specific. When the APIs take more than 5 params and only differ in one or two, it can get a little difficult to navigate. :-)

sample-compose/src/main/java/com/airbnb/lottie/sample/compose/examples/AnimatableExamplesPage.kt
gpeal: @doris4lt did reading the rest of the examples clarify this? The LottieAnimation overload handles this case.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
gpeal: @doris4lt any suggestions on the name?

lottie-compose/src/main/java/com/airbnb/lottie/compose/animateLottieComposition.kt
gpeal: @doris4lt It's effectively the declarative version of LottieAnimatable

lottie-compose/src/main/java/com/airbnb/lottie/compose/lottieTransition.kt
```diff
@@ -0,0 +1,166 @@
...
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.collectLatest
+
+/**
+ * [lottieTransition] allows you to define individual animations mapped to a set of states.
+ *
+ * Use this version if your transition is defined by separate segments from within a single composition
+ *
+ * Inside of [animate], you will probably want to use the suspending version of [animateLottieComposition].
```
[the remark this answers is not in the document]
gpeal: Gone and replaced by the functions on LottieAnimatable.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimatable.kt
```diff
@@ -0,0 +1,261 @@
...
+     * @param clipSpec An optional [LottieClipSpec] to trim the playback of the composition between two values.
+     * @param initialProgress An optional progress value that the animation should start at. Defaults to the
+     *                        starting progress as defined by the clipSpec and speed. Can be used to resume
+     *                        animations from arbitrary points.
+     * @param cancellationBehavior The behavior that this animation should have when cancelled. In most cases,
+     *                             you will want it to cancel immediately. However, if you have a state based
+     *                             transition and you want an animation to finish playing before moving on to
+     *                             the next one then you may want to set this to [LottieCancellationBehavior.OnFinish].
+     */
+    suspend fun animate(
```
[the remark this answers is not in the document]
gpeal: @doris4ly. I actually had this in some of the iterations on this branch. One of them copied over the values if `continueFromPreviousAnimate` was true. I could see this going either way and am open to a) default values, b) retained values c) retained of continue is set to true.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
```diff
@@ -10,142 +11,226 @@ import androidx.compose.ui.graphics.drawscope.drawIntoCanvas
...
-    animationState: LottieAnimationState = rememberLottieAnimationState(autoPlay = true),
-) {
-    LottieAnimation(compositionResult(), animationState, modifier)
-}
-
 @Composable
 fun LottieAnimation(
     composition: LottieComposition?,
-    state: LottieAnimationState,
+    @FloatRange(from = 0.0, to = 1.0) progress: Float,
```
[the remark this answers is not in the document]
gpeal: Yes! That's exactly why I'm excited about this! Fwiw, it was possible with the previous version of this PR too though.

sample-compose/src/main/java/com/airbnb/lottie/sample/compose/examples/AnimatableExamplesPage.kt
doris4lt: I get why this is needed. It is a 3-step setup that I recommend slimming down a bit for better ergonomic. 

Maybe consider combining the first two steps with:  `val (composition, anim) = lottieComposition(...)`

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimatable.kt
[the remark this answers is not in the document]
doris4lt: The best way to work this out is to write some user code for a couple of real-world use cases, and see which option requires the least amount of effort. This comment was just about ergonomics.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
doris4lt: ProgressBasedLottie? ProgressDrivenLottie? LottieByProgress? :D Something that implies: 1) it doesn't animate on its own and 2) driven by a progress would add clarity.

lottie-compose/src/main/java/com/airbnb/lottie/compose/animateLottieComposition.kt
doris4lt: When would a user use it?

lottie-compose/src/main/java/com/airbnb/lottie/compose/lottieTransition.kt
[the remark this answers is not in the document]
doris4lt: In that case, you might want to update this kdoc.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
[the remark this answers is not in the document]
doris4lt: 🥇

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimatable.kt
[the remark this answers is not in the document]
gpeal: @doris4lt that's that's the purpose of the Examples pages, actually. This version is a little more verbose but I made the change because in one of my examples, it wound up doing the wrong thing. I'll play around with this a little more though :smile:

lottie-compose/src/main/java/com/airbnb/lottie/compose/lottieTransition.kt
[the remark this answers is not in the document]
gpeal: @dori4lt, yup, thank you!

lottie-compose/src/main/java/com/airbnb/lottie/compose/animateLottieComposition.kt
gpeal: @doris4lt I think _most_ of the time. 90%+ maybe? The imperative version (LottieAnimatable) should be relatively rare IMO. Maybe this should be called `animateLottieCompisitionAsState` to match Compose `animate*asState`. Same analogy here.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
gpeal: Naming is hard :smile: All Lottie animations are driven by progress though. The difference is whether it's internal or external. I think I can remove the overload that takes the spec directly which would reduce this to two overloads. Do you think that would sufficiently help?

sample-compose/src/main/java/com/airbnb/lottie/sample/compose/examples/AnimatableExamplesPage.kt
gpeal: @dors4lt I think 90% of people will be using the declarative `LottieAnimataion` composable. `LottieAnimatable` is also not coupled to a single composition so I don't think it makes sense to combine the two here. You can see one such example on the transitions examples page.

lottie-compose/src/main/java/com/airbnb/lottie/compose/animateLottieComposition.kt
doris4lt: When should devs use this:

```
@Composable
fun LottieAnimation(
    composition: LottieComposition?,
    modifier: Modifier = Modifier,
    isPlaying: Boolean = true,
    restartOnPlay: Boolean = true,
    clipSpec: LottieClipSpec? = null,
    speed: Float = 1f,
    iterations: Int = 1,
    imageAssetsFolder: String? = null,
    imageAssetDelegate: ImageAssetDelegate? = null,
    outlineMasksAndMattes: Boolean = false,
    applyOpacityToLayers: Boolean = false,
    enableMergePaths: Boolean = false,
) 
```
over `animateLottieComposition` + `LottieAnimation`?

If the difference between the two is just whether progress is observable, it may be worth merging the two options above by making the progress always observable and externally immutable, which can be achieved by returning LottieAnimationState.

The benefit of doing this is to reduce selection anxiety when devs are faced with a number of API options. :P

gpeal: It's partly that and partly that one is declarative and one is imperative. I like having the two options because the declarative version doesn't require 2 setup lines of code. You can handle the entire thing in just composable parameters.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
gpeal: I removed the extra overloads so there are only 2 now and I think it makes sense for them to have the same name. I honestly can't think of a better name here 😄

PR #1846, opened 2021-07-15
Add cancellation behavior to animateLottieCompositionAsState

headsvk:
I added the option to set cancellation behavior on `animateLottieCompositionAsState` instead of relying on the default value from `LottieAnimatable.animate`. It's a pretty useful setting that allows me to easily implement my loading animation that includes success and failure transitions.


lottie-compose/src/main/java/com/airbnb/lottie/compose/animateLottieCompositionAsState.kt
```diff
@@ -36,6 +36,7 @@ fun animateLottieCompositionAsState(
     clipSpec: LottieClipSpec? = null,
     speed: Float = 1f,
     iterations: Int = 1,
+    cancellationBehavior: LottieCancellationBehavior = LottieCancellationBehavior.Immediately,
```
gpeal, line 39: Can you add a param kdoc?

gpeal (commented):
Neat, thanks!

lottie-compose/src/main/java/com/airbnb/lottie/compose/animateLottieCompositionAsState.kt
headsvk: ah right, I added the one from `animate`

gpeal (approved):

PR #1851, opened 2021-07-18
Fix Loader Animation

sidharthv96:
Initial loader would stop after a Single iteration.
Changed to Infinite.


sample-compose/src/main/java/com/airbnb/lottie/sample/compose/composables/Loader.kt
```diff
@@ -18,6 +19,7 @@ fun Loader(
     LottieAnimation(
         composition,
         modifier = modifier
-            .size(100.dp)
+            .size(128.dp),
+        iterations = LottieConstants.IterateForever
```
gpeal, line 23: Thanks for fixing this but modifier should always be the last parameter.

sidharthv96: Thanks for the review. First time working with compose, wasn't aware of the conventions. 
I had to delete the fork, I'll close this and open a fresh one with iterations above modifier.

PR #1877, opened 2021-08-25
Optimize gradle & Remove redundant .gitignore files

Goooler:
- Add `base.gradle` & `kotlin.gradle` for modules to apply and simplifying boilerplate.
- Remove redundant .gitignore files.
- Extrace version strings to ext block.
- Add arg to [Optimizing Gradle builds for JDK 11](https://developer.android.com/studio/releases/gradle-plugin#optimize-gc-jdk-11).
- Add [Dependabot](https://dependabot.com/) to automated dependency updates.


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
```
gpeal, line 1: I would rather not add dependabot. It will actually add more work for me to manage the pull requests as a single maintainer.

gradle/base.gradle
```diff
@@ -0,0 +1,18 @@
+android {
```
gpeal, line 1: Can you undo this change? I'd rather keep the build files self-contained. There aren't that many modules and I would like each one, especially the ones that get published, to be very explicit.

Goooler: I can revert them. 
But I think I just extracted common parts of modules' gradle, and properties could be overrided.

.github/dependabot.yml
Goooler: Removed.

build.gradle
```diff
@@ -2,9 +2,20 @@ import org.ajoberstar.grgit.Grgit
...
+    lifecycleVersion = '2.3.1'
+    composeVersion = '1.0.2'
     kotlinVersion = '1.5.21'
     daggerVersion = '2.38.1'
+    awsVersion = '2.8.3'
+    mockitoVersion = '3.12.4'
+    robolectricVersion = '4.6.1'
+    retrofitVersion = '2.9.0'
+    materialVersion = '1.4.0'
+    ePoxyVersion = '4.6.3'
```
gpeal, line 18:
```suggestion
    epoxyVersion = '4.6.3'
```

gradle/base.gradle
gpeal: Please revert them

sample-compose/build.gradle
```diff
@@ -1,23 +1,18 @@
 plugins {
   id 'com.android.application'
-  id 'kotlin-android'
+  id "kotlin-android"
```
gpeal, line 3: Any reason you changed just this one?

gpeal (approved):

sample-compose/build.gradle
Goooler: No, just effected by reverting.

PR #1885, opened 2021-09-01
Downgrade espresso and AWS

gpeal:
https://github.com/airbnb/lottie-android/pull/1874/ broke instrumentation tests:
1) Espresso broke with [this erro](https://github.com/airbnb/lottie-android/pull/1874/). It can probably be fixed by upgrading androidx test core but then there is a dependency conflict error which can probably be fixed by upgrading `androidx.fragment:fragment-testing` but there isn't a version of it that depends on test core 1.4.0.
2) The upgraded AWS SDK returns multiple status from the listener which caused the await continuation to be already resumed. I don't have time to debug the full behavior change and the old one was working fine.

### DIFF
--- a/sample-compose/build.gradle
+++ b/sample-compose/build.gradle
@@ -83,5 +83,5 @@ dependencies {
 
   testImplementation 'junit:junit:4.13.2'
   androidTestImplementation 'androidx.test.ext:junit:1.1.3'
-  androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
+  androidTestImplementation 'androidx.test.espresso:espresso-core:3.3.0'
 }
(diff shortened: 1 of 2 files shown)


spades67:
> https://github.com/airbnb/lottie-android/pull/1874/ broke instrumentation tests:
> 1) Espresso broke with [this erro](https://github.com/airbnb/lottie-android/pull/1874/). It can probably be fixed by upgrading androidx test core but then there is a dependency conflict error which can probably be fixed by upgrading `androidx.fragment:fragment-testing` but there isn't a version of it that depends on test core 1.4.0.
> 2) The upgraded AWS SDK returns multiple status from the listener which caused the await continuation to be already resumed. I don't have time to debug the full behavior change and the old one was working fine.

.github/workflows/validate.yml

PR #1897, opened 2021-09-10
Enforce order of operations to avoid rounding errors

scottdweber:
Previously, the calculation of `endLength` could suffer from rounding errors. While most of the time, these errors are not a big deal, when the end value is 100, this can result in `endLength` being larger than `totalLength`, causing unexpected behavior where the animation reverts to an initial state when it should be showing the final state. By forcing the order of operations, we can get the answer we expect / need.

For example:

```kotlin
val totalLength = 98.164566
totalLength * 100f / 100f // 98.16457
totalLength * (100f / 100f) // 98.164566
```

Unfortunately, I have been unable to get a simple animation to reproduce this error, and the complex one I am using is not really available for public sharing. However, it seems like this should be a safe change.


scottdweber:
Oh, nice! Those visual tests show an example of the bug and that it is fixed with this PR 🕺

gpeal (approved):
Thanks for catching and fixing this!

PR #1915, opened 2021-10-04
Don't cache dpScale in Utils

eartle:
It's possible for the screen density to change when moving between windows or screens. In this case lottie renders animations the wrong size. Android already does a pretty good job of caching and updating the system context so just let it do it's job.

I am currently relying on reflection to clear this field on a configuration change and I'd of course like to remove that.

I haven't had time to come up with a generic repro, but I thought I'd open this anyway. Sorry.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
@@ -66,7 +66,6 @@ public final class Utils {
   };
 
   private static final float INV_SQRT_2 = (float) (Math.sqrt(2) / 2.0);
-  private static float dpScale = -1;
 
   private Utils() {
   }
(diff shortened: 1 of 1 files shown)


gpeal:
@eartle why does screen density change in between screens?

eartle:
@gpeal because different screens may have different densities. You can set this up in an emulator here

Screenshot 2021-10-04 at 15 29 37

eartle:
And here is the bug. Open keyboard on first screen looks good.

Screenshot 2021-10-04 at 15 39 28

Open keyboard on second screen and you'll notice the two lottie animation views on the keyboard toolbar (the i and the three dots) are drawn the wrong size.

Screenshot 2021-10-04 at 15 41 09

gpeal:
@eartle Thanks for that. I'll have to benchmark this function on real devices because it might get called hundreds or thousands of times. It might be necessary to do something slightly smarter here.

This also won't solve all cases because a lot of the dp conversions are done at parse time.

eartle:
@gpeal We're also calling `LottieCompositionFactory.clearCache` on configuration change which should hopefully deal with everything else. From code inspection is looks like the Context and DisplayMetrics are already cached so I'm hopefully about the benchmarking, but you never know. Thanks!

Mbrittny (approved):

Mbrittny (commented):
[]()

gpeal (approved):

Issue #1921, opened 2021-10-12
Caused by: java.lang.StringIndexOutOfBoundsException: length=0; index=0 in lottie library


RohanBhagat1:
Hi, I am facing an issue with the Lottie animation.
The json file I am trying to render in the android app is not rendering. The stack trace I am getting is added below.
Though the json file is running appropriately on Lottie site.

StackTrace:
2021-10-12 14:56:15.392 11174-11174/in.ludo.supremegold E/AndroidRuntime: FATAL EXCEPTION: main
    Process: in.ludo.supremegold, PID: 11174
    java.lang.IllegalStateException: Unable to parse composition
        at com.airbnb.lottie.LottieAnimationView$1.onResult(LottieAnimationView.java:78)
        at com.airbnb.lottie.LottieAnimationView$1.onResult(LottieAnimationView.java:71)
        at com.airbnb.lottie.LottieAnimationView$3.onResult(LottieAnimationView.java:95)
        at com.airbnb.lottie.LottieAnimationView$3.onResult(LottieAnimationView.java:88)
        at com.airbnb.lottie.LottieTask.notifyFailureListeners(LottieTask.java:158)
        at com.airbnb.lottie.LottieTask.access$200(LottieTask.java:27)
        at com.airbnb.lottie.LottieTask$1.run(LottieTask.java:133)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:241)
        at android.app.ActivityThread.main(ActivityThread.java:7582)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:941)
     Caused by: java.lang.StringIndexOutOfBoundsException: length=0; index=0
        at java.lang.String.charAt(Native Method)
        at android.graphics.Color.parseColor(Color.java:1384)
        at com.airbnb.lottie.parser.LayerParser.parse(LayerParser.java:129)
        at com.airbnb.lottie.parser.LottieCompositionMoshiParser.parseLayers(LottieCompositionMoshiParser.java:121)
        at com.airbnb.lottie.parser.LottieCompositionMoshiParser.parse(LottieCompositionMoshiParser.java:87)
[...]

PR #1924, opened 2021-10-14
The issue #1921 is related to the problem with the json file itself. …

klaus19:
Fixes issue #1921

I created a Demo App and tried to display this JSON file and it gave me similar type of error mentioned above so I used another Lottie file and they worked fine. This the GitHub link of Demo App [https://github.com/klaus19/LottieError] that I have created which shows how the error is actually in the JSON file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 import org.ajoberstar.grgit.Grgit
 
 buildscript {
+  ext.kotlin_version = '1.6.0-RC'
   ext {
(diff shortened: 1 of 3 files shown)


klaus19:
@gpeal what do you mean by PR, I have actually implemented the issue on my demo app and found the error.

gpeal:
@klaus19 These changes are not needed in the build.gradle files here. I'm still not exactly clear on what this has to do with #1921 but if you want to comment there, feel free.

PR #1927, opened 2021-10-18
Removed unnecessary int cast to avoid Integer allocations

### LINKED ISSUE #1926: Unnecessary allocations in ColorKeyframeAnimation.getIntValue()
**Link to fork with a repro in the issue-repro module**
https://github.com/pablorengo/lottie-android

**Describe the bug**
Several Integer allocations are being made for every frame of an animation that changes its color using a ValueCallback. This is making the GC to execute really often, having impact in app performance.

**Steps To Reproduce**

1. Start the issue-repro app
2. Go to Profiler
3. Select the Memory profiler
4. Start Java/Kotlin allocation recording
5. After a few seconds you will see a huge amount of Integer allocations.

image

pablorengo:
This pull request attempts to reduce Integer allocations when using an animation that has a ValueCallback to change some colors.

Resolves #1926


gpeal:
@pablorengo I'm a bit surprised that this actually eliminates the auto-boxing. Were you able to verify that this achieves that?

PR #1929, opened 2021-10-20
Replaced Paint.setAlpha() with Paint.setColor() to reduce unnecessary allocations

### LINKED ISSUE #1928: Calling Paint.setAlpha() generates several ColorSpace$Named[] allocations
**Link to fork with a repro in the issue-repro module**
https://github.com/pablorengo/lottie-android

**Describe the bug**
Calling an Enum `values()` method clones the values of the enum, allocating memory. If this is executed in a loop, we can see a significant amount of allocations in a short period of time. Calling Paint.setAlpha() generates a ColorSpace which calls Named.values() where Named is an Enum. This was modified in Android 30.

**Steps To Reproduce**

1. Start the issue-repro app
2. Go to Profiler
3. Select the Memory profiler
4. Start Java/Kotlin allocation recording
5. After a few seconds you will see a huge amount of ColorSpace$Named[] allocations.

image

image

pablorengo:
Resolves #1928 

This pull request attempts to remove ColorSpace.Named[] allocations when calling Paint.setAlpha(). These changes will fix the issue only when using RGB colors.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/animation/LPaint.java
+++ b/lottie/src/main/java/com/airbnb/lottie/animation/LPaint.java
@@ -1,8 +1,11 @@
 package com.airbnb.lottie.animation;
 
+import static com.airbnb.lottie.utils.MiscUtils.clamp;
+
 import android.graphics.Paint;
 import android.graphics.PorterDuff;
 import android.graphics.PorterDuffXfermode;
+import android.os.Build;
 import android.os.LocaleList;
 
 import androidx.annotation.NonNull;
@@ -34,5 +37,14 @@ public class LPaint extends Paint {
   @Override
   public void setTextLocales(@NonNull LocaleList locales) {
     // Do nothing.
+  }
+
+  @Override public void setAlpha(int alpha) {
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+      int color = getColor();
+      setColor((clamp(alpha, 0, 255) << 24) | (color & 0xFFFFFF));
+    } else {
+      super.setAlpha(clamp(alpha, 0, 255));
+    }
   }
 }
--- a/lottie/src/main/java/com/airbnb/lottie/animation/LPaint.java
+++ b/lottie/src/main/java/com/airbnb/lottie/animation/LPaint.java
@@ -39,7 +39,15 @@ public class LPaint extends Paint {
     // Do nothing.
   }
 
-  @Override public void setAlpha(int alpha) {
+  /**
+   * Overrides {@link android.graphics.Paint#setAlpha(int)} to avoid
+   * unnecessary {@link android.graphics.ColorSpace.Named ColorSpace$Named[] }
+   * allocations when calling this method in Android 29 or lower.
+   *
+   * @param alpha set the alpha component [0..255] of the paint's color.
+   */
+  @Override
+  public void setAlpha(int alpha) {
     if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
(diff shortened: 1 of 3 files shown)


gpeal:
@pablorengo I'm not seeing the same behavior as you. I tried your repro project (thank you for that, by the way) on API 30 and 31 and am not seeing any allocations other than the autoboxing Integer for the dynamic properties.

[This](https://cs.android.com/android/platform/superproject/+/android-11.0.0_r48:frameworks/base/graphics/java/android/graphics/ColorSpace.java;l=213;drc=android-11.0.0_r48) seems to be the `ColorSpace.values()` call you're referring to but it is only called once ever so I'm not sure where the behavior you're seeing is happening.

pablorengo:
@gpeal Sorry, I forgot to mention it here too. This problem happens on API level <= 29. It was fixed on Android 30 code.

gpeal:
@pablorengo Lottie uses `LPaint()` which already has an optimization for text locale. This could be upstreamed to that so that it applies everywhere, not jus for strokes and fills.

pablorengo:
@gpeal Great. Let me check it and I'll come back with the modification. Thanks.

gpeal:
@pablorengo Are you still interested in doing this?

pablorengo:
@gpeal Yes, I'll try to finish it today

lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
```diff
@@ -98,9 +98,9 @@ public FillContent(final LottieDrawable lottieDrawable, BaseLayer layer, ShapeFi
       return;
     }
     L.beginSection("FillContent#draw");
-    paint.setColor(((ColorKeyframeAnimation) colorAnimation).getIntValue());
+    int color = ((ColorKeyframeAnimation) this.colorAnimation).getIntValue();
     int alpha = (int) ((parentAlpha / 255f * opacityAnimation.getValue() / 100f) * 255);
-    paint.setAlpha(clamp(alpha, 0, 255));
+    paint.setColor((clamp(alpha, 0, 255) << 24) | (color & 0xFFFFFF));
```
gpeal, line 103: @pablorengo You can revert this now right?

pablorengo: @gpeal In this case, it avoids calling `setColor()` twice.

lottie/src/main/java/com/airbnb/lottie/animation/LPaint.java
```diff
@@ -35,4 +38,13 @@ public LPaint(int flags, PorterDuff.Mode porterDuffMode) {
   public void setTextLocales(@NonNull LocaleList locales) {
     // Do nothing.
   }
+
+  @Override public void setAlpha(int alpha) {
```
gpeal, line 42: Could you add a doc explaining why this is necessary so future people can understand?

lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
gpeal: Ah, I see. Got it. We could also make `setColorAndAlpha` on `LPaint` in case there are other usages of this in the future

lottie/src/main/java/com/airbnb/lottie/animation/LPaint.java
gpeal: It's not a "memory issue". It's avoiding memory allocations.

gpeal (approved):

PR #1931, opened 2021-10-22
Provide a second overload to TextDelegate.getText that provides layerName

### LINKED ISSUE #1932: Add Keypath to TextDelegate in order to have same cross platform behavior as iOS AnimationTextProvider.

**If you don't use this template, your issue will be closed. Delete this text once completed.**

Currently if an animation is shared across Android and iOS the text layers are intercepted by their respective libraries differently. Android (TextDelegate) can only replace text with `text` that is named X, whereas iOS (AnimationTextProvider) will be able to look at either the `Keypath` and the `text`, by default (DictionaryTextProvider) will look for matching KeyPath that is named X. This creates a bit of a disconnect but prevents duplicate text among layers in an animation as well for android. Consider the following example set of layers

    layer_1 with text "READ IT" // Let's say in context it's a commanding statement
    layer_2 with text "READ IT" // Let's say in context it's a statement representing the past

By default the layers would be named the same as the text they contained. However this is just default. In the case where a content author would want to disambiguate the layers they can rename them. In this case, Android would not be able to replace the text on layer1 and layer2 independently. They would both be replaced with the same exact string. If TextDelegate were to be using to localize the text, it might be incorrect as text here could mean different things depending on context.

I proposed #1931 as a possible solution. It merely exposes the `KeyPath`'s name to the interface of `TextDelegate` to make an implementation possible that would address the above example. 

We could also generate multiple animations that are organized specifically for Android vs iOS but that may defeat the overall purpose of the portable animation format. It's the current solution that we are going with but it's not ideal and very error prone.

**Additional context**

greggiacovelli:
In the iOS version of lottie, the [AnimationTextProvider](https://github.com/airbnb/lottie-ios/blob/master/lottie-swift/src/Public/TextProvider/AnimationTextProvider.swift) can look up text with the context of the keypath and the text within the layer. This doesn't seem to be possible in the Android version. 

Closes #1932 
Given the following example though there doesn't seem to be a way to address the following pseudo animation the same way across platforms.

    layer_1 with text "READ IT" // Let's say in context it's a commanding statement
    layer_2 with text "READ IT" // Let's say in context it's a statement representing the past

The default implementations aside ([DictionaryTextProvider](https://github.com/airbnb/lottie-ios/blob/master/lottie-swift/src/Public/TextProvider/AnimationTextProvider.swift#L18) and [TextDelegate](https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/TextDelegate.java)), if an implementation was needed to disambiguate strings that were duplicated inside an animation, this would not be possible in the Android library today simply because `getText(String input)` is all that is provided. This just expands getText to also have access to the `keyPath`. 

I am not sure if this has already been covered so please refer me to such past decisions if this is an inappropriate pull request.


lottie/src/main/java/com/airbnb/lottie/model/layer/TextLayer.java
```diff
@@ -245,7 +246,7 @@ private void drawTextWithFont(
     String text = documentData.text;
     TextDelegate textDelegate = lottieDrawable.getTextDelegate();
     if (textDelegate != null) {
-      text = textDelegate.getTextInternal(text);
+      text = textDelegate.getTextInternal(getName(), text);
```
gpeal, line 249: This isn't a key path name. This is the layerName. Key path would be the fully qualified list of elements up to the root composition

greggiacovelli: I see that, I if I am understanding this is similar to iOS though looking at the implementation. It is only the layer name.

greggiacovelli: For more context, this is what I am referring to.

https://github.com/airbnb/lottie-ios/blob/64564c12ba0c258f5a5d9b1e81455b1f1da2315d/lottie-swift/src/Private/LayerContainers/CompLayers/CompositionLayer.swift#L63

    CompositionLayer {
      init(layer: LayerModel, size: CGSize) {
          ...
          self.keypathName = layer.name
          ...
      }
    }

I didn't think this was the true keypath either, but just a string representing the layer name.

gpeal: @greggiacovelli Interesting. It does seem like iOS is using the name slightly inconsistently. For the same of consistency within the Android library, I'd prefer `layerName` here

greggiacovelli: Ok will do. Thanks for the guidance and I will append the PR in a bit

greggiacovelli: addressed in 2099504f

gpeal (approved):
Thanks!

Issue #1387, opened 2019-09-12
Crash: java.lang.IllegalStateException Unable to create layer for LottieAnimationView, size 1088x-448 max size 8192 color type 4 has context 1 


lchad:
I just upgrade lottie version from 2.7.0 to 3.0.7.
I got some new crash in the latest version app.

StackTrace in Android 9:
# main(2)

java.lang.IllegalStateException

Unable to create layer for LottieAnimationView, size 1088x-448 max size 16384 color type 4 has context 1

1 android.os.MessageQueue.nativePollOnce(Native Method)
--
2 android.os.MessageQueue.next(MessageQueue.java:330)
3 android.os.Looper.loop(Looper.java:169)
4 android.app.ActivityThread.main(ActivityThread.java:7083)
5 java.lang.reflect.Method.invoke(Native Method)
6 com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
7 com.android.internal.os.ZygoteInit.main(ZygoteInit.java:876)

Anyone is try to fix this problem?

GeeJoe:
Face the same issue，when will this problem to be solved?

lchad:
it's is really hard to reproduce.
in a one million DAU App, crash about 10 times everyday.

Sanjay-F:
+1 

---

**StackTrace in Android 5.0/5.1:**
  
java.lang.IllegalStateException: Unable to create layer for LottieAnimationView @62b16e5
at android.os.MessageQueue.nativePollOnce(Native Method)
at android.os.MessageQueue.next(MessageQueue.java:143)
at android.os.Looper.loop(Looper.java:122)
at android.app.ActivityThread.main(ActivityThread.java:5418)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1037)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:832)

GeeJoe:
> it is really hard to reproduce.
> in a one million DAU App, crash about 10 times everyday.

+1 @gpeal @lchad

gpeal:
@GeeJoe @Sanjay-F You can use `setRenderMode(RenderModel.SOFTWARE)` to get around this.

DineshBabuYadav:
[...]

PR #1952, opened 2021-11-28
Replace LottieAnimationView layer types with internal bitmap rendering

gpeal:
This change marks a significant change in the underlying rendering pipeline for Lottie.

Previously, Lottie would always set a layerType on LottieAnimationView. For hardware accelerated animations, this meant that a separate graphics layer was allocated for the animation. The disadvantage with this method is that the texture has to be uploaded separately to the GPU on each frame.

For software rendering, Lottie would depend on Android's view caching mechanism in which View allocates a drawing cache bitmap which it then draws to a canvas. This has the disadvantage of the Bitmap always being the size of the containing View even if it's larger than the drawable.
It also abstracts away the Bitmap so that further optimizations. When software rendering is enabled and it has to be redrawn without getting invalidated then it can skip re-rendering and simply redraw the cached bitmap.
It also upstream the bitmap rendering form LottieAnimationView to LottieDrawable which allows it to be exposed to lottie-compose which is important for the same reasons it is important for the base library and wasn't possible before.

While working on this PR, I tried rendering and then drawing only the bounds of the animation. However, I found that calculating the bounds for the entire animation was slower than drawing the entire bitmap (which is very fast).

Fixes https://github.com/airbnb/lottie-android/issues/1387

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -1,14 +1,11 @@
 package com.airbnb.lottie;
 
-import static com.airbnb.lottie.RenderMode.HARDWARE;
-
 import android.animation.Animator;
 import android.animation.ValueAnimator;
 import android.content.Context;
 import android.content.res.ColorStateList;
 import android.content.res.TypedArray;
 import android.graphics.Bitmap;
-import android.graphics.Canvas;
 import android.graphics.ColorFilter;
 import android.graphics.drawable.Drawable;
 import android.os.Build;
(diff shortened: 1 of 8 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -1127,46 +1146,13 @@ public void setApplyingOpacityToLayersEnabled(boolean isApplyingOpacityToLayersE
...
-   * Disable the extraScale mode in {@link #draw(Canvas)} function when scaleType is FitXY. It doesn't affect the rendering with other scaleTypes.
-   *
-   * <p>When there are 2 animation layout side by side, the default extra scale mode might leave 1 pixel not drawn between 2 animation, and
-   * disabling the extraScale mode can fix this problem</p>
-   *
-   * <b>Attention:</b> Disable the extra scale mode can downgrade the performance and may lead to larger memory footprint. Please only disable this
-   * mode when using animation with a reasonable dimension (smaller than screen size).
+   * This API no longer has any effect.
    */
+  @Deprecated
```
sanogueralorenzo, line 1151: How do you normally handle this? When it comes a major version you will remove it from the codebase?

sanogueralorenzo (approved):

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
gpeal: @sanogueralorenzo I don't think I ever have before but probably as you said.

PR #1965, opened 2021-12-15
Reproduce dynamic property bug

DougSig:


gpeal:
Thanks for creating this, it was helpful to reproduce it quickly. I'm going to close this since it won't land but will put up a fix.

PR #2005, opened 2022-01-21
Add a benchmark that can generate Baseline Profiles for Lottie.

tikurahul:
## Add a benchmark that can generate Baseline Profiles for Lottie.

* Add `LottieBaselineProfiles.kt` that can generate Baseline Profiles.
* Add `LottieBaselineBenchmark.kt` to benchmark the effect of Baseline Profiles.
* Add `baseline-prof.txt` files for `lottie` and `lottie/compose`.

## Benchmark Results (Release builds)

```
LottieBaselineBenchmark_benchmarkNoCompilation
  timeToInitialDisplayMs   min   507.4,   median   507.4,   max   507.4
          frameCpuTimeMs   P50     16.8,   P90     31.2,   P95     52.6,   P99    122.4
    Traces: Iteration 0

LottieBaselineBenchmark_benchmarkBaselineProfiles
  timeToInitialDisplayMs   min 417.0,   median 417.0,   max 417.0
          frameCpuTimeMs   P50   16.3,   P90   21.5,   P95   22.6,   P99   44.7
    Traces: Iteration 0
```

## Regenerate Baseline Profiles

To regenerate baseline profiles, just re-run the test defined in
`LottieBaselineProfiles.kt`. This should produce an output file with
the optimal baseline profile for the app defined in `sample-compose`.

To extract the relevant portions, use the tool defined [here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:development/splitBaselineProfiles).
This tool takes 2 command line arguments:

`profilePath` - This is the path to the baseline profile generated by the test.
`checkoutPath` - This is the path to the folder that contains `settings.gradle` (the module root).

This tool generates an output file, that splits the baseline profile
based on the module layout defined in the checkout path.


gpeal (approved):
Thanks for submitting this! What should the recommendation here be for people who ship apps with lottie-compose? Can Lottie publish a baseline similar to proguard rules or is this simply something that will speed up the sample apps in this repo?

tikurahul:
Apps that want to use Lottie or Lottie Compose don't need to do anything. Libraries with a `baseline-prof.txt` participate in the build via a process similar to manifest merging. 

So these rules get merged with other rules defined, and make apps faster without any additional work on their part (as long as they use a recent version of AGP).

gpeal:
@tikurahul Perfect, that's what I was hoping to here!

keyboardsurfer:
I've seen that the Baseline Profile has not been updated since the original contribution in January last year.
Lottie has seen a major version bump from 5 to 6 in the meantime. 

With the [Baseline Profile Gradle plugin](https://developer.android.com/topic/performance/baselineprofiles/create-baselineprofile#baseline-profile-gradle-plugin) you could automatically generate a new Baseline Profile for every release.

gpeal:
Thanks, @keyboardsurfer! I would happily accept a PR for this or I'll create a new issue to track it and hopefully get around to it soonish

PR #2015, opened 2022-02-11
Add id duplicated check when parsing json

zaydenyang:
…cessary.

Sample file: https://lottiefiles.com/95828-loadinganimation，it preview well on website but not on android devices. The layer's id who's named "BigBall" and "SmallBall" is duplicated with others, cause their ShapeLayer's parentLayer is null and finally effect the matrix when draw.


Mr-lidajun:
这是来自我的假期自动回复邮件。您好，我最近正在“夏威夷”，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #2016, opened 2022-02-11
Add id duplicated check when parsing json and tips if necessary.

zaydenyang:
Sample file: https://lottiefiles.com/95828-loadinganimation , it preview well on website but not on android devices. The layer's id who's named "BigBall" and "SmallBall" is duplicated with others, cause their ShapeLayer's parentLayer is null and finally effect the matrix when draw.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/parser/LottieCompositionMoshiParser.java
+++ b/lottie/src/main/java/com/airbnb/lottie/parser/LottieCompositionMoshiParser.java
@@ -89,7 +89,7 @@ public class LottieCompositionMoshiParser {
           parseAssets(reader, composition, precomps, images);
           break;
         case 8:
-          parseFonts(reader, fonts);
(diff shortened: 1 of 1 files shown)


gpeal:
This is actually not because of the duplicate names. It is because "ind" is the same for multiple layers in the same compositiob.

I believe this is a bug with bodymovin or the json was hand-modified. There should not be duplicate indices. You can change the "ind" property by hand and then this will render just fine.

zaydenyang:
> This is actually not because of the duplicate names. It is because "ind" is the same for multiple layers in the same compositiob.
> 
> I believe this is a bug with bodymovin or the json was hand-modified. There should not be duplicate indices. You can change the "ind" property by hand and then this will render just fine.

Is it a better choice to present friendly tip when 'ind' duplicates by the bug of bodyMovie or manual change ?

PR #2028, opened 2022-03-02
Use the Apache License Url in the pom

### LINKED ISSUE #2027: Use Standard POM License
We are using Licensee to validate that we don't accidentally sneak licenses into our project:
https://github.com/cashapp/licensee

This does not work with lottie because it uses a custom path for the license:

https://repo1.maven.org/maven2/com/airbnb/android/lottie/5.0.2/lottie-5.0.2.pom

```xml
<license>
<name>Apache License 2.0</name>
<url>
https://github.com/airbnb/lottie-android/blob/master/LICENSE
</url>
<distribution>repo</distribution>
</license>
```

It would be great if you published your pom linked to the offiical apache license url, like for example in:

https://repo1.maven.org/maven2/com/squareup/leakcanary/plumber-android/2.8.1/plumber-android-2.8.1.pom
```xml
<license>
<name>The Apache Software License, Version 2.0</name>
<url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
<distribution>repo</distribution>
</license>
```

PaulWoitaschek:
Fixes #2027


gradle.properties
```diff
@@ -6,8 +6,8 @@ POM_URL=https://github.com/airbnb/lottie-android
 POM_SCM_URL=https://github.com/airbnb/lottie-android
 POM_SCM_CONNECTION=scm:git@github.com:airbnb/lottie-android.git
 POM_SCM_DEV_CONNECTION=scm:git@github.com:airbnb/lottie-android.git
-POM_LICENSE_NAME=Apache License 2.0
-POM_LICENSE_URL=https://github.com/airbnb/lottie-android/blob/master/LICENSE
+POM_LICENSE_NAME=Apache-2.0
```
gpeal, line 9: I don't think this change is "more correct". Searching through GitHub, I see lots of projects that use `Apache license 2.0` and almost none that use `Apache-2.0`. `The Apache Software License, Version 2.0` actually seems to be the most common.

gradle.properties
```diff
@@ -6,8 +6,8 @@ POM_URL=https://github.com/airbnb/lottie-android
...
+POM_LICENSE_URL=https://www.apache.org/licenses/LICENSE-2.0.txt
```
gpeal, line 10: Licensee should support this. This is a common pattern. I've never use it myself but could you double check?

gradle.properties
JakeWharton: Per https://maven.apache.org/pom.html#Licenses this change is bringing it in line with the new recommendations. Specifically,

> Using an [SPDX identifier](https://spdx.org/licenses/) as the license name is recommended.

gradle.properties
JakeWharton: The canonical Apache 2 URL is specified by [the SPDX database](https://spdx.org/licenses/Apache-2.0.html) as https://www.apache.org/licenses/LICENSE-2.0, but https://www.apache.org/licenses/LICENSE-2.0.txt is explicitly supported as a fallback URL: https://github.com/cashapp/licensee/blob/9281c96d83d664ec8c39ba0a0adac3bf7bfccc44/src/main/kotlin/app/cash/licensee/licenses.kt#L64

gpeal: Thanks for the info, @JakeWharton. So is the recommended approach the full proposal here minus `.txt`?

JakeWharton: I would feel more confident saying yes if the POM format had a statement that using the canonical SPDX URL was recommended, but both of these Apache URLs are so ubiquitous that I don't think it matters all that much.

PaulWoitaschek: I changed it to the non txt one as the apache license states:

> You may obtain a copy of the License at
> 
> http://www.apache.org/licenses/LICENSE-2.0

They really should recommend using https for obtaining their license 🙄

gradle.properties
```diff
@@ -6,8 +6,8 @@ POM_URL=https://github.com/airbnb/lottie-android
 POM_SCM_URL=https://github.com/airbnb/lottie-android
 POM_SCM_CONNECTION=scm:git@github.com:airbnb/lottie-android.git
 POM_SCM_DEV_CONNECTION=scm:git@github.com:airbnb/lottie-android.git
-POM_LICENSE_NAME=Apache License 2.0
-POM_LICENSE_URL=https://github.com/airbnb/lottie-android/blob/master/LICENSE
+POM_LICENSE_NAME=Apache-2.0
+POM_LICENSE_URL=https://www.apache.org/licenses/LICENSE-2.0
```
gpeal, line 10: I think we should delete the LICENSE in the repo now, right?

JakeWharton: Per https://www.apache.org/licenses/LICENSE-2.0 you are recommended to distribute a copy.

> ## How to apply the Apache License to your work
> 
> Include a copy of the Apache License, typically in a file called LICENSE, in your work

gpeal (approved):
Thanks for doing this @PaulWoitaschek and thanks for the guidance @JakeWharton!

Issue #2069, opened 2022-05-02
LottieAnimationView doesn't restore the animation progress


flaringapp:
**Describe the bug**

Since [v5.0.1](https://github.com/airbnb/lottie-android/releases/tag/v5.0.1) `LottieAnimationView` has a simple logic to prevent `onRestoreInstanceState` from overriding user-provided values. Those changes were introduced in [this PR](https://github.com/airbnb/lottie-android/pull/2002).
So the issue is about `progress` that won't be restored if user sets a value before state restoration - same logic as for other parameters. However, the progress is always set in the view `init` function called from a constructor. Therefore, progress restoration is always ignored.

Progress setter:
```
public void setProgress(@FloatRange(from = 0f, to = 1f) float progress) {
  userActionsTaken.add(UserActionTaken.SET_PROGRESS);
  lottieDrawable.setProgress(progress);
}
```
Progress restoration:
```
if (!userActionsTaken.contains(UserActionTaken.SET_PROGRESS)) {
  setProgress(ss.progress);
}
```
Setting progress in the `init` function:
```
setImageAssetsFolder(ta.getString(R.styleable.LottieAnimationView_lottie_imageAssetsFolder));
setProgress(ta.getFloat(R.styleable.LottieAnimationView_lottie_progress, 0));
enableMergePathsForKitKatAndAbove(ta.getBoolean(
    R.styleable.LottieAnimationView_lottie_enableMergePathsForKitKatAndAbove, false));
```

**Steps To Reproduce**

1. Open Fragment A with `LottieAnimationView` proving any animation (raw resource)
2. Play animation (doesn't matter if it finishes or not)
3. Replace Fragment A with Fragment B (Fragment A saves it's state)
4. Go back from Fragment B to Fragment A (Fragment A restores it's state)
5. Animation progress is not restored

gpeal:
Would you like to put up a PR for this?

flaringapp:
Sure. How about preventing `userActionsTaken.add(UserActionTaken.SET_PROGRESS)` from being called if `setProgress()` is called from the constructor?

gpeal:
@flaringapp I'd have to look into it in more detail but seems like a reasonable start.

ketanMuttha:
We are also facing a similar issue. is there any resolution?

PR #2072, opened 2022-05-10
Fix LottieAnimationView.setProgress called from the view init method prevents the progress from being restored from the saved state

flaringapp:
See [#2069](https://github.com/airbnb/lottie-android/issues/2069)

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -182,7 +182,7 @@ import java.util.Set;
     setImageAssetsFolder(ta.getString(R.styleable.LottieAnimationView_lottie_imageAssetsFolder));
 
     boolean hasProgress = ta.hasValue(R.styleable.LottieAnimationView_lottie_progress);
-    setProgressInternal(hasProgress, ta.getFloat(R.styleable.LottieAnimationView_lottie_progress, 0f));
+    setProgressInternal(ta.getFloat(R.styleable.LottieAnimationView_lottie_progress, 0f), hasProgress);
 
     enableMergePathsForKitKatAndAbove(ta.getBoolean(
         R.styleable.LottieAnimationView_lottie_enableMergePathsForKitKatAndAbove, false));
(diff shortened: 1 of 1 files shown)


gpeal:
If you were to set an initial progress but then were to subsequently change it or play the animation, wouldn't restoration overwrite the saved progress with what is set in XML?

flaringapp:
@gpeal I think it makes sense to check whether there is a progress attribute set in the XML and if yes - treat it as user-provided and prevent saved state restoration from overwriting the progress value.

Also it will be nice to provide an option to restore saved progress even if the progress is set in the XML. I suggest adding an attribute named `lottie_useRestoredProgress` with default value `false`. So if user wants to explicitly restore progress state instead of starting the animation from the specified progress, he can use this attribute. Behind the scenes it'll affect the `setProgressInternal` called from attributes init method. If the attribute is `true`, then we won't save the `UserActionTaken.SET_PROGRESS` flag (`fromUser = false`) or do exactly the opposite with default `false` value

gpeal:
@flaringapp I'm still trying to wrap my head around the exact use case that's failing here. Could you create a sample project that highlights the issue?

flaringapp:
@gpeal Please take a look at this [sample project](https://github.com/flaringapp/LottieIssue)
In scope of this issue I expect the animation to persist it's previous progress when a fragment is restored

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -977,7 +980,15 @@ public int getFrame() {
   }
 
   public void setProgress(@FloatRange(from = 0f, to = 1f) float progress) {
-    userActionsTaken.add(UserActionTaken.SET_PROGRESS);
+    setProgressInternal(true, progress);
+  }
+
+  private void setProgressInternal(
```
gpeal, line 986: Could you swap the parameter order here?

gpeal (commented):
Okay thanks for the sample. I've been very busy lately but found some time to take a look at this. let's move forward with your implementation as-is. Just swap the parameter order here but other than that, LGTM. Thanks for doing this.

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
flaringapp: @gpeal Done

flaringapp: I've rebased on master as well

gpeal (approved):

PR #2078, opened 2022-05-19
Fix LottieAnimation recomposes on every frame degrading performance

### LINKED ISSUE #2077: Compose: LottieAnimation recomposes on every frame degrading performance
Issue Repro Compose Fork : [https://github.com/MSDarwish2000/lottie-android-performance](Fork)

**Describe the bug**
`LottieAnimation` recomposes itself on every frame. This is completely unnecessary in most cases as the only actual change happens during the draw phase, not composition or layout phases. This happens as a result of passing `progress` to `LottieAnimation`. Even when using the function without progress, this happens internally reducing the scope of recomposition but not preventing it.

**Steps To Reproduce**

1. This can be reproduced with any Lottie Compose sample using Layout Inspector in Android Studio Dolphin
2. This can be benchmarked using the code in the issue above or any similar code

**Sample benchmark result**

- _While running minified release variant_
 Drawing time: 1174.3ms - Recomposition time: 979.4ms - percentage: 45.48%
- _While running debug variant_
 Drawing time: 1369.3ms - Recomposition time: 1219ms - percentage: 47.1%
- _While debugging_
 Drawing time: 2398.3ms - Recomposition time: 1837.2ms - percentage: 43.38%

_Note:_ The percentage of recomposition time may increase if the parent composable had more code which is very likely

MSDarwish2000:
Fixes #2077

This issue can be fixed by a very simple approach mentioned in [https://developer.android.com/jetpack/compose/performance#defer-reads](Compose performance docs - Defer reads as long as possible) by deferring the progress read to the draw function.

Also, I've updated samples in this repo to use the new approach.

### DIFF
--- a/issue-repro-compose/src/main/java/com/airbnb/lottie/issues/compose/ComposeIssueReproActivity.kt
+++ b/issue-repro-compose/src/main/java/com/airbnb/lottie/issues/compose/ComposeIssueReproActivity.kt
@@ -22,6 +22,6 @@ class ComposeIssueReproActivity : AppCompatActivity() {
     fun Content() {
         val composition by rememberLottieComposition(LottieCompositionSpec.RawRes(R.raw.heart))
         val progress by animateLottieCompositionAsState(composition)
-        LottieAnimation(composition, progress)
+        LottieAnimation(composition, { progress })
     }
 }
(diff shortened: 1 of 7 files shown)


gpeal (approved):
Thanks for doing this. The API is a bit unfortunate but you're correct about the approach here. I'm going to land this and then do some small tweaks to APIs/docs and publish a release.
I'll leave the overload there but deprecated for one release and then will remove it.

MSDarwish2000:
@gpeal Thank you for merging the pull request and maintaining this awesome project.

PR #2081, opened 2022-05-26
De-dupe gradient stops

gpeal:
Refer to code comments for more info:
Before:
CleanShot 2022-05-26 at 11 25 30
After:
CleanShot 2022-05-26 at 11 26 54

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
+++ b/lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
@@ -234,34 +234,22 @@ public class GradientColorParser implements com.airbnb.lottie.parser.ValueParser
 
     int aIndex = 0;
     int bIndex = 0;
-    int mergedIndex = 0;
     int numDuplicates = 0;
     // This will be the merged list but may be longer than what is needed if there are duplicates.
     // If there are, the 0 elements at the end need to be truncated.
     float[] mergedNotTruncated = new float[arrayA.length + arrayB.length];
     for (int i = 0; i < mergedNotTruncated.length; i++) {
-      final float a;
-      if (aIndex < arrayA.length) {
-        a = arrayA[aIndex];
-      } else {
-        a = Float.NaN;
-      }
-
-      final float b;
-      if (bIndex < arrayB.length) {
-        b = arrayB[bIndex];
-      } else {
-        b = Float.NaN;
-      }
+      final float a = aIndex < arrayA.length ? arrayA[aIndex] : Float.NaN;
+      final float b = bIndex < arrayB.length ? arrayB[bIndex] : Float.NaN;
 
       if (Float.isNaN(b) || a < b) {
-        mergedNotTruncated[mergedIndex++] = a;
+        mergedNotTruncated[i] = a;
         aIndex++;
       } else if (Float.isNaN(a) || b < a) {
-        mergedNotTruncated[mergedIndex++] = b;
+        mergedNotTruncated[i] = b;
         bIndex++;
       } else {
-        mergedNotTruncated[mergedIndex++] = a;
+        mergedNotTruncated[i] = a;
         aIndex++;
         bIndex++;
         numDuplicates++;
@@ -273,8 +261,6 @@ public class GradientColorParser implements com.airbnb.lottie.parser.ValueParser
     }
 
 
-    float[] merged = new float[mergedNotTruncated.length - numDuplicates];
-    System.arraycopy(mergedNotTruncated, 0, merged, 0, merged.length);
-    return merged;
+    return Arrays.copyOf(mergedNotTruncated, mergedNotTruncated.length - numDuplicates);
   }
 }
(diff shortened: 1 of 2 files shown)


lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
```diff
@@ -223,4 +221,60 @@ private int getColorInBetweenOpacityStops(float position, int color, float[] opa
...
+    }
+
+    if (numDuplicates == 0) {
+      return mergedNotTruncated;
+    }
+
+
+    float[] merged = new float[mergedNotTruncated.length - numDuplicates];
+    System.arraycopy(mergedNotTruncated, 0, merged, 0, merged.length);
+    return merged;
```
denis-bezrukov, line 278: nit: 
```suggestion
    return Arrays.copyOf(mergedNotTruncated, mergedNotTruncated.lengh - numDuplicates);
```

lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
```diff
@@ -223,4 +221,60 @@ private int getColorInBetweenOpacityStops(float position, int color, float[] opa
...
+
+      final float b;
+      if (bIndex < arrayB.length) {
+        b = arrayB[bIndex];
+      } else {
+        b = Float.NaN;
+      }
+
+      if (Float.isNaN(b) || a < b) {
+        mergedNotTruncated[mergedIndex++] = a;
```
denis-bezrukov, line 258: isn't `mergeIndex` can be replaced by `i`?

lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
```diff
@@ -223,4 +221,60 @@ private int getColorInBetweenOpacityStops(float position, int color, float[] opa
...
+    // This will be the merged list but may be longer than what is needed if there are duplicates.
+    // If there are, the 0 elements at the end need to be truncated.
+    float[] mergedNotTruncated = new float[arrayA.length + arrayB.length];
+    for (int i = 0; i < mergedNotTruncated.length; i++) {
+      final float a;
+      if (aIndex < arrayA.length) {
+        a = arrayA[aIndex];
+      } else {
+        a = Float.NaN;
+      }
```
denis-bezrukov, line 248: nit could be elvis

```suggestion
      final float a = aIndex < arrayA.length ? arrayA[aIndex] : Float.NaN
```

denis-bezrukov (approved):

lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
gpeal: Done!

lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
gpeal: Yup!

lottie/src/main/java/com/airbnb/lottie/parser/GradientColorParser.java
gpeal: 👍

Vodanhlatui:
> Tham khảo các nhận xét về mã để biết thêm thông tin: Trước: Sau: CleanShot 2022-05-26 lúc 11 25 30 CleanShot 2022-05-26 lúc 11 26 54

imbryk:
@gpeal @denis-bezrukov 
This PR is crashing the app for quite a few complex animations we use.
I'm not able to identify the place in our animations, as they are really big, but I've found that the problem is that this PR  changes the length of color array in `GradientColor` which results in https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/animation/keyframe/GradientColorKeyframeAnimation.java being created with keyframes which have GradientColors of different lengths, which eventually crashes at the `GradientColor.lerp` method with `java.lang.ArrayIndexOutOfBoundsException`

This can theoretically be simply fixed not to crash, but I believe that would completely break the intended gradient animation.

If the problem solved by this PR was needed mostly for static lottie files, maybe you could apply this change only when parsing a static value:
https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/parser/KeyframeParser.java#L369
but when parsing animation keep the old way of parsing it (i.e. keeping all values)

imbryk:
@gpeal would it be possible to send it to you privately?

Blincwear88:
#2297

gpeal:
@imbryk Yeah, go ahead and DM it to me on Twitter but open a new issue saying you did that so it doesn't slip through the cracks the next time I work on this.

imbryk:
@gpeal I am not able to send you a message on twitter 🤷 - please grab the files from gdrive:
https://drive.google.com/drive/folders/1LOiSaIJq1LJL848Ne8rA9Omrn2ErynfY

PR #2103, opened 2022-06-28
Add fontStyle argument for FontAssetDelegate

jhen0409:
In my case, using FontAssetDelegate to listen `fetchFont(fontFamily)` was not enough to know what font to load.

For example, I got `fontFamily == "Roboto"` but don't know the font style is `Black`. This PR will solve this problem.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/FontAssetDelegate.java
+++ b/lottie/src/main/java/com/airbnb/lottie/FontAssetDelegate.java
@@ -11,16 +11,16 @@ import android.graphics.Typeface;
 @SuppressWarnings({"unused", "WeakerAccess"}) public class FontAssetDelegate {
 
   /**
-   * Override this if you want to return a Typeface from a font family.
+   * Override this if you want to return a Typeface from a font family and style.
    */
-  public Typeface fetchFont(String fontFamily) {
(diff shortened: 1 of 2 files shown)


gpeal:
Could you test https://github.com/airbnb/lottie-android/blob/master/snapshot-tests/src/main/assets/Tests/EmbeddedFont.zip? It looks like, from the snapshot tests, that the embedded font is no longer loading in this PR.

jhen0409:
I did some testing on local. I have no permission to use that Happo / AWS S3 feature, so I replaced some s3 upload logic on debug.

- Currently I can't successfully render text, and the snapshot file size always 0 on local. Maybe have some guide for how to run snapshot-tests correctly?
- It seems does get it's own font (by [`Font.getTypeface()`](https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/manager/FontAssetManager.java#L90-L92)), no different with master.

jhen0409:
Tested on sample (add EmbeddedFont.zip to assets). It seems works as expected (Pixel 6):

Screenshot (Jun 30, 2022 12_20_27 PM)

gpeal (approved):

PR #2115, opened 2022-08-01
Fix ImageAssetManager.bitmapForId NPE crash

hsrzq:
In method ImageAssetManager.bitmapForId, if BitmapFactory.decodeStream returned null, so NPE crash happened. And we can see this is a legal return value in android docs, So I fixed it.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
+++ b/lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
@@ -121,6 +121,10 @@ public class ImageAssetManager {
       Logger.warning("Unable to decode image.", e);
       return null;
     }
+    if (bitmap == null) {
+        Logger.warning("Decoded image is NULL.");
+        return null;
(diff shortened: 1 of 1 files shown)


lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
```diff
@@ -121,6 +121,7 @@ public void setDelegate(@Nullable ImageAssetDelegate assetDelegate) {
       Logger.warning("Unable to decode image.", e);
       return null;
     }
+    if (bitmap == null) return null;
```
gpeal, line 124: Can you add a warning here just like the IllegalArgumentException branch above?

hsrzq: Yes, It's necessary.

hsrzq:
This is my first PR, I think I made some mistake, so I closed this PR and create a new one.

PR #2117, opened 2022-08-08
Fix ImageAssetManager.bitmapForId NPE crash

hsrzq:
In Android docs, BitmapFactory.decodeStream may return NULL, This is a legal return value, we did not deal with it, NPE may occur in some cases. So I fixed it.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
+++ b/lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
@@ -122,8 +122,8 @@ public class ImageAssetManager {
       return null;
     }
     if (bitmap == null) {
-        Logger.warning("Decoded image is NULL.");
-        return null;
+      Logger.warning("Decoded image `" + id + "` is null.");
+      return null;
     }
     bitmap = Utils.resizeBitmapIfNeeded(bitmap, asset.getWidth(), asset.getHeight());
     return putBitmap(id, bitmap);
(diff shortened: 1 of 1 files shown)


lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
```diff
@@ -121,6 +121,10 @@ public void setDelegate(@Nullable ImageAssetDelegate assetDelegate) {
       Logger.warning("Unable to decode image.", e);
       return null;
     }
+    if (bitmap == null) {
+        Logger.warning("Decoded image is NULL.");
```
gpeal, line 125:
```suggestion
        Logger.warning("Decoded image (${id}) is null.");
```

Do you mind changing the `Unable to decode image.` above to `Unable to decode image(${id}).` as well? I think it will be more informative.

hsrzq: Very good advice. But this is Java not Kotlin, so I made a little change.

lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java
```diff
@@ -121,6 +121,10 @@ public void setDelegate(@Nullable ImageAssetDelegate assetDelegate) {
       Logger.warning("Unable to decode image.", e);
       return null;
     }
+    if (bitmap == null) {
+      Logger.warning("Decoded image `" + id + "` is null.");
```
gpeal, line 125: Could you add the same message on line 121?

hsrzq: OK

gpeal (approved):

Issue #2125, opened 2022-09-01
LottieAnimation Composable: support LottieAnimationView (XML) repeatMode equivalent


brentwatson:
`lottie-compose`'s `LottieAnimation` Composable does not support the equivalent of `LottieAnimationView` `app:lottie_repeatMode=reverse|restart` XML attribute.

It would be great if `LottieAnimation` were to add something like a `repeatMode: LottieRepeatMode` parameter.

I've been able to do this manually by switching between a positive and negative animation speeds. But it would be nice if this was supported internally by `LottieAnimation`.

```kotlin
@Composable
fun LottieAnimationLoopReverse(
    modifier: Modifier = Modifier,
    asset: LottieCompositionSpec.Asset,
) {
    val composition by rememberLottieComposition(asset)
    var shouldReverse by remember { mutableStateOf(false) }
    val lottieAnimation = rememberLottieAnimatable()
    if (shouldReverse) {
        LaunchedEffect(composition) {
            lottieAnimation.animate(composition = composition, speed = -1f)
            shouldReverse = false
        }
    } else {
        LaunchedEffect(composition) {
            lottieAnimation.animate(composition = composition, speed = 1f)
            shouldReverse = true
        }        
    }

    LottieAnimation(
        composition = composition,
        progress = lottieAnimation.value,
        modifier = modifier,
    )
}
```

PR #2128, opened 2022-09-14
Add reverseOnRepeat option to LottieAnimatable

brentwatson:
Resolves issue #2125 - `LottieAnimation Composable: support LottieAnimationView (XML) repeatMode equivalent`.

This PR adds a `reverseOnRepeat` boolean option to `LottieAnimatable` (defaults to `false`). When set to true the effect is the same as using `app:lottie_repeatMode="reverse"` in the Android XML lottie implementation.

#### Regular Repeat:

https://user-images.githubusercontent.com/1331002/190251017-31ca9346-f009-4b4c-bf90-c8a615778ddc.mov

#### With reverseOnRepeat=true:

https://user-images.githubusercontent.com/1331002/190251270-52c9bc98-8937-42a2-a84a-b1295c604e08.mov


lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimatable.kt
```diff
@@ -160,12 +161,19 @@ private class LottieAnimatableImpl : LottieAnimatable {
...
+    override var reverseOnRepeat: Boolean by mutableStateOf(false)
+        private set
+
     override var clipSpec: LottieClipSpec? by mutableStateOf(null)
         private set
 
     override var speed: Float by mutableStateOf(1f)
         private set
 
+    private val speedValue: Float by derivedStateOf {
```
gpeal, line 173: Does this need to be a property or can it be derived as a local variable on line 291?

brentwatson: Done - 71154c638ff8d8d7bdc1863cef5a90f042575c6c

The value must remain a `derivedStateOf` instance otherwise the animation misbehaves. If you think that adds too much additional complexity to `onFrame` I can revert the change.

Thanks for the review!

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimatable.kt
```diff
@@ -278,9 +284,12 @@ private class LottieAnimatableImpl : LottieAnimatable {
         val minProgress = clipSpec?.getMinProgress(composition) ?: 0f
         val maxProgress = clipSpec?.getMaxProgress(composition) ?: 1f
 
-        val dProgress = dNanos / 1_000_000 / composition.duration * speed
+        val speedValue: Float by derivedStateOf {
```
gpeal, line 287: What is the derivedStateOf doing here? This isn't a composable function.

brentwatson: See my comment above -- https://github.com/airbnb/lottie-android/pull/2128#discussion_r975522959

I took a closer look at why making `speedValue` a val on L287 doesn't work as intended. The reason is that the rest of `onFrame` needs an updated `speedValue` each time `iteration` changes. So we would need a `var` on L287 and a _reassigment_ on L306. 

Based on this I decided that might further complicate `onFrame` which is already fairly complex so I reverted back to using a property for `speedValue`. If you would prefer the var + reassignment instead let me know.

lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimatable.kt
brentwatson: If `speedValue` is changed to a local `val` which isn't updated on L306 then the animation misbehaves and looks like this...

https://user-images.githubusercontent.com/1331002/194068499-bea1e6cc-ee9c-4606-9692-0dd6bc731658.mov

If you would prefer a `var` which is set on L287 and reassigned within `onFrame` on L306 I can make that change, it just seemed to complicate the function IMO instead of reading from a `derivedStateOf` property which does the calculation for you.

gpeal: Do you know why it's not working when it's a local val?

gpeal: I see. Could you add some documentation for this property since `speed` and `speedValue` is ambiguous on its own. It's not much better but maybe `rawSpeed` or `actualSpeed` or `speedWithRepeatMode`?

brentwatson: Good idea. I added some docs to the property to explain why & when a negative value can be returned. I opted for `frameSpeed` for the property since that's more descriptive of how it is used. I didn't want to change `speed` -> `rawSpeed` as that would effect that public API of the class (as where the new `frameSpeed` property is private).

08f69299c99a1ecd2e99a51e548941bcb00ee62c

gpeal (approved):

gpeal:
Thanks for your contribution!

agustinsivoplas:
I am trying to use reverseOnRepeat but is not working for me. Do you have an example? My idea is pretty simple. Is a checkbox that user can check or uncheck on click and the animation should be reversed on click. Any help?

Thanks in advance

brentwatson:
reverseOnRepeat only makes sense with `repeat=true`. If you want to play an
animation once in reverse try setting `speed=-1` then back to `speed=1` to
play if forward again.

PR #2156, opened 2022-11-02
png optimization(loss less)

pea-sys:
Lighter png files. Image quality is not degraded.
I used WinMerge when checking the image loss less


gpeal (commented):
Sorry, I can see that you're just going through lots of public repos and generating these PRs. I'm only accepting PRs that were specifically designed for Lottie.

Issue #2104, opened 2022-06-29
LottieAnimationView crashes in 5.2.0


giangpham96:
After updating the library to 5.2.0, my app starts crashing for animations with gradients. I suspect Lottie adds some extra verifications in this version, but not sure what is wrong with the lottie file. Version 5.1.1 works fine

Here is the stacktrace
```
Fatal Exception: java.lang.IllegalArgumentException: Cannot interpolate between gradients. Lengths vary (8 vs 10)
       at com.airbnb.lottie.model.content.GradientColor.lerp(GradientColor.java:30)
       at com.airbnb.lottie.animation.keyframe.GradientColorKeyframeAnimation.getValue(GradientColorKeyframeAnimation.java:19)
       at com.airbnb.lottie.animation.keyframe.GradientColorKeyframeAnimation.getValue(GradientColorKeyframeAnimation.java:8)
       at com.airbnb.lottie.animation.keyframe.BaseKeyframeAnimation.getValue(BaseKeyframeAnimation.java:144)
       at com.airbnb.lottie.animation.content.GradientFillContent.getRadialGradient(GradientFillContent.java:206)
       at com.airbnb.lottie.animation.content.GradientFillContent.draw(GradientFillContent.java:131)
       at com.airbnb.lottie.animation.content.ContentGroup.draw(ContentGroup.java:185)
       at com.airbnb.lottie.animation.content.ContentGroup.draw(ContentGroup.java:185)
       at com.airbnb.lottie.model.layer.ShapeLayer.drawLayer(ShapeLayer.java:36)
       at com.airbnb.lottie.model.layer.BaseLayer.draw(BaseLayer.java:292)
       at com.airbnb.lottie.model.layer.CompositionLayer.drawLayer(CompositionLayer.java:128)
       at com.airbnb.lottie.model.layer.BaseLayer.draw(BaseLayer.java:250)
       at com.airbnb.lottie.LottieDrawable.drawDirectlyToCanvas(LottieDrawable.java:1354)
       at com.airbnb.lottie.LottieDrawable.draw(LottieDrawable.java:517)
```

Here is the lottie file
```json
[...]

PR #2160, opened 2022-11-11
Allow interpolating between gradients with different opacity stops

gpeal:
Fixes https://github.com/airbnb/lottie-android/issues/2104


param-aggarwal-groww:
How to check which version of the library has this fix?

PR #2210, opened 2022-12-19
Fix AnimatorPauseListener callbacks

Viacheslav01:
LottieValueAnimator doesn't utilize a base implementation of pause and resume methods of Animator.
The result is undispatched events to AnimatorPauseListener.


Viacheslav01:
How can I run snapshot-tests correctly?

gpeal:
It won't work locally without the api key either 🤔 I just changed how CI works (to make it faster) but will need to think about how to best handle this.

Viacheslav01:
Is this PR has a chance to be merged?

gpeal:
Could you try updating this branch from master? The snapshot tests should work now.

gpeal (approved):

gpeal:
Thanks for adding this!

gpeal (commented):
Actually, before landing this, could you add the APIs to LottieDrawable and LottieAnimationView (with javadocs) just like the other listeners?

https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java#L844

Viacheslav01:
It was implemented in PR https://github.com/airbnb/lottie-android/commit/70fe2f89fa118045fdade501c289bd6182742ee9 but it didn't work properly without a correct listeners handling.

PR #2225, opened 2023-01-17
Update text rendering y position

JonathonFry:
This PR updates text positioning to use the box position y value (`ps`) 

This already works on iOS, I think this is because it [uses the position object](https://github.com/airbnb/lottie-ios/blob/167812e8ce59698144714b81a4132d3d188f2a39/Sources/Private/MainThread/LayerContainers/CompLayers/TextCompositionLayer.swift#L141) directly. 

**Example**

Before|After
--|--
lottie-example-before|lottie-text-example-after


gpeal:
Thank! Could you add an animation that repros this to [this folder](https://github.com/airbnb/lottie-android/tree/master/snapshot-tests/src/main/assets/Tests). Then it'll show up in the snapshot diffs.

JonathonFry:
Will do - just getting some design support to export a small sample file to reproduce 🙇

PR #2253, opened 2023-03-01
Fixed the stream was not closed when the lottie animation hit the cache

sunlocker:
This PR fixes the stream not closing when the lottie animation hits the cache.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -130,7 +130,7 @@ public class LottieCompositionFactory {
         LottieCompositionCache.getInstance().put(cacheKey, result.getValue());
       }
       return result;
-    });
+    }, null);
   }
 
   /**
@@ -184,7 +184,7 @@ public class LottieCompositionFactory {
   public static LottieTask<LottieComposition> fromAsset(Context context, final String fileName, @Nullable final String cacheKey) {
     // Prevent accidentally leaking an Activity.
     final Context appContext = context.getApplicationContext();
-    return cache(cacheKey, () -> fromAssetSync(appContext, fileName, cacheKey));
+    return cache(cacheKey, () -> fromAssetSync(appContext, fileName, cacheKey), null);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


sachin2dehury (approved):

lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -311,7 +311,7 @@ private static boolean isNightMode(Context context) {
    * @see #fromJsonInputStreamSync(InputStream, String, boolean)
    */
   public static LottieTask<LottieComposition> fromJsonInputStream(final InputStream stream, @Nullable final String cacheKey) {
-    return cache(cacheKey, () -> fromJsonInputStreamSync(stream, cacheKey));
+    return cache(cacheKey, () -> fromJsonInputStreamSync(stream, cacheKey), () -> closeQuietly(stream));
```
gpeal, line 314: `fromJsonInputStreamSync` closes the stream [here](https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java#L332). Is that not sufficient for your use case?

sunlocker: Thank you for reviewing.

Yes, `fromJsonInputStreamSync` doesn't close my stream because I set stream and cacheKey via `setAnimation(InputStream stream, @Nullable String cacheKey)` and hit the cache, but `fromJsonInputStreamSync` doesn't execute when there is cache.

I also can't close my stream manually because the Lottie framework internally closes my stream via `fromJsonInputStreamSync` when there is no cache.

ankit-gautam23 (commented):
I don't think closing stream is needed it comes under capabilities of Lottie framework. It can do it by itself when cache is not there

sunlocker:
@ankit-gautam23 Yes, I also think that the Lotti framework should close the stream, but the framework now does not close the stream when there is a cache

ankit-gautam23:
Thanks for the Clarification @sunlocker

gpeal (approved):
Thanks for doing this and sorry it took so long to review.

PR #2286, opened 2023-05-07
Allow for non-closing input stream parsing

### LINKED ISSUE #2252: Request LottieCompositionFactory fromXXXInputStream close param to be public.
I use `Lottie` in `Glide`, and decode it in `ResourceDecoder` which has `InputStream`, when I use `LottieCompositionFactory.fromJsonInputStreamSync` and `LottieCompositionFactory.fromZipStreamSync`, they all close the `InputStream`, but I don't want `LottieCompositionFactory` close the `InputStream`.

I think, make the private `LottieCompositionFactory.fromJsonReaderSyncInternal` and `LottieCompositionFactory.fromZipStreamSyncInternal` to be public will be better.

gpeal:
Fixes #2252

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -315,6 +315,17 @@ public class LottieCompositionFactory {
   }
 
   /**
+   * @see #fromJsonInputStreamSync(InputStream, String, boolean)
+   */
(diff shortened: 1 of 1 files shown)


peerless2012:
Zip InputStream still not support control close or not.

gpeal:
Is that something you're using as well?

peerless2012:
Yes,  ZipInputStream need controll close for user.

PR #2316, opened 2023-06-14
Bump dependencies, remove all deprecations, improve CI builds

mateuszkwiecinski:
Another set of build updates. I'll try to comment out each of them. And again, a reminder, please let me know if these are not aligned with the direction this project heads to :) 

I have a couple more ideas of things that could be improved. A few examples:
 - introducing static code analysis for Kotlin files
 - running more checks on the CI (building sample apps, running UI tests)
 - Sample app, the one available on Google Play, does not work properly when scanning QR codes
 - Improve testability by fixing the re-initialization (one of the opened non-rendering issues)

 and maybe, eventually, once I get more familiar with how Lottie library works internally, I'll be able to fix the bug reported if it is even _fixable_ 😅


.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+indent_size = 4
+ij_kotlin_allow_trailing_comma = true
+ij_kotlin_allow_trailing_comma_on_call_site = true
+
+[*.java]
+ij_java_class_count_to_use_import_on_demand = 3
+ij_java_names_count_to_use_import_on_demand = 3
+ij_java_packages_to_use_import_on_demand = 3
+ij_java_blank_lines_after_class_header = 1
+ij_java_imports_layout = *,|,javax.**,java.**,|,$*
```
mateuszkwiecinski, line 17: Ah, I just realized this project has codestyle file committed, but my IDE still used the _default_ one :/ 
I thought there is no _suggested_ code style so I thought I should add one, to avoid unnecessary changes in the imports. 

I can remove the file, although I'd strongly suggest slowly migrating towards IDE-agnostic `.editorconfig`, and leave at least the upper part, up to kotlin codestyle (which I believe will be needed if you decide to introduce static code analysis).

.github/workflows/validate.yml
```diff
@@ -21,8 +21,8 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'zulu'
-          java-version: 17
-          cache: 'gradle'
+          java-version: 20
+      - uses: gradle/gradle-build-action@v2
```
mateuszkwiecinski, line 25: From my experience, the official `gradle-build-action` from Gradle do the job better than what `setup-java` does. + it also comes with extra summary that simplifies reading build data from build scans:

image

the improvement won't be immediately visible, since it works in read-only mode when invokend from PR, but things will start working normally after it's ivoken from the main branch for the first time

.github/workflows/validate.yml
```diff
@@ -21,8 +21,8 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'zulu'
-          java-version: 17
-          cache: 'gradle'
+          java-version: 20
```
mateuszkwiecinski, line 24: Since the target java version is properly set-up and the library will always be compatible down to java-8, it's safe to benefit from latest build improvements available in latest java version. 

This change is inspired by https://jakewharton.com/build-on-latest-java-test-through-lowest-java/

If the _policy_ is to use LTS version for some reason, I can easily revert that

lottie-compose/src/test/java/com/airbnb/lottie/compose/LottieAnimatableImplTest.kt
```diff
@@ -431,11 +430,7 @@ class LottieAnimatableImplTest {
         assertEquals("lastFrameNanos at %d".format(frameTimeMs), lastFrameNanos, anim.lastFrameNanos)
     }
 
-    private fun runTest(test: suspend CoroutineScope.() -> Unit) {
-        runBlockingTest {
-            withContext(clock) {
-                test()
-            }
-        }
+    private fun runTest(test: suspend CoroutineScope.() -> Unit) = kotlinx.coroutines.test.runTest(context = clock + UnconfinedTestDispatcher()) {
```
mateuszkwiecinski, line 433: Source: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/MIGRATION.md#replace-runblockingtest-with-runtestunconfinedtestdispatcher

lottie/build.gradle
```diff
@@ -41,7 +41,6 @@ dependencies {
 
   annotationProcessor libs.nullaway
   errorprone libs.errorprone.core
-  errorproneJavac libs.errorprone.javac
```
mateuszkwiecinski, line 44: AFAIU this was needed only when running on Java 8

sample-compose/src/androidTest/java/com/airbnb/lottie/samples/InfiniteAnimationTest.kt
```diff
@@ -23,7 +24,7 @@ import org.junit.Test
 class InfiniteAnimationTest {
 
     @get:Rule
-    val composeTestRule = createAndroidComposeRule(ComposeActivity::class.java)
+    val composeTestRule = createAndroidComposeRule(ComponentActivity::class.java)
```
mateuszkwiecinski, line 27: These tests were failing for me, with an error saying `setContent` can't be called twice. 
I noticed you're overwriting Content, so I used empty _generic_ activity

build.gradle
```diff
@@ -12,7 +12,7 @@ plugins {
 }
 
 allprojects {
-  def compileJavaVersion = JavaVersion.VERSION_17
+  def compileJavaVersion = JavaVersion.VERSION_20
```
gpeal, line 15: Let's stick with 17

https://developer.android.com/build/releases/gradle-plugin#jdk-17-agp

.github/workflows/validate.yml
gpeal, line 24:
```suggestion
          java-version: 17
```
Let's stick with 17

versions.properties
```diff
@@ -130,483 +102,30 @@ version.io.jsonwebtoken..jjwt=0.9.1
...
-##                  # available=1.8.0-alpha03
-##                  # available=1.8.0-beta01
-##                  # available=1.8.0-rc01
-##                  # available=1.8.0
-##                  # available=1.9.0-alpha01
-##                  # available=1.9.0-alpha02
-##                  # available=1.9.0-beta01
-##                  # available=1.9.0-rc01
-##                  # available=1.9.0
+version.google.android.material=1.9.0
```
mateuszkwiecinski, line 105: In my $workProject, minor version bumps of material library usually cause screenshot-tests to fail 🤷 I'm not yet sure what's covered, but I won't be surprised if this change fails some screenshots

lottie-compose/src/test/java/com/airbnb/lottie/compose/LottieAnimatableImplTest.kt
gpeal, line 433:
```suggestion
    private fun runTest(test: suspend CoroutineScope.() -> Unit) = runTest(context = clock + UnconfinedTestDispatcher()) {
```

build.gradle
```diff
@@ -38,6 +38,10 @@ allprojects {
         sourceCompatibility targetJavaVersion
         targetCompatibility targetJavaVersion
       }
+
+      tasks.withType(JavaCompile).configureEach {
+        options.compilerArgs += ["-Xlint:-options"] // Remove when targeting higher Java version
```
gpeal, line 43: What is this for?

.editorconfig
```diff
@@ -0,0 +1,17 @@
+root = true
```
gpeal, line 1: What reads .editorconfig?

gpeal (commented):
So far these cleanups have been helpful. Thanks for taking the time to do this. If you ever want to ask about a change in advance, feel free to shoot me a DM on Twitter @gpeal8

versions.properties
```diff
@@ -130,483 +102,30 @@ version.io.jsonwebtoken..jjwt=0.9.1
...
-##                            # available=2.62.2
-##                            # available=2.63.0
-##                            # available=2.64.0
-##                            # available=2.65.0
-##                            # available=2.66.0
-##                            # available=2.67.0
-##                            # available=2.68.0
-##                            # available=2.69.0
-##                            # available=2.70.0
+version.com.amazonaws..aws-android-sdk-s3=2.71.0
```
mateuszkwiecinski, line 124: FYI, there is a deprecation warning, which I didn't have a chance to fix:

image

I don't have a possibility to run these tests locally :/

versions.properties
```diff
@@ -989,97 +267,27 @@ version.androidx.benchmark=1.1.0-beta01
...
 ##            # available=1.8.0-alpha03
 ##            # available=1.8.0-alpha04
+##            # available=1.8.0-alpha05
 
-plugin.net.ltgt.errorprone=3.0.1
-##             # available=3.1.0
+plugin.net.ltgt.errorprone=3.1.0
 
 plugin.com.google.devtools.ksp=1.8.21-1.0.11
+##                 # available=1.8.22-1.0.11
```
mateuszkwiecinski, line 286: Compose is not yet compatible with `1.8.22` so I kept kotlin & ksp at 1.8.21

versions.properties
```diff
@@ -621,33 +140,13 @@ version.com.airbnb.android..mvrx=1.5.1
...
-##                        # available=2.5.0
-##                        # available=2.5.1
-##                        # available=2.6.0
-##                        # available=2.6.1
-##                        # available=2.7.0
-##                        # available=3.0.0
-##                        # available=3.0.1
-##                        # available=3.0.2
-##                        # available=3.0.3
+version.com.airbnb.android..mavericks=3.0.3
```
mateuszkwiecinski, line 145: I'm not familiar with the library at all 😬 I didn't know what to expect, but it seems like everything works like expected. I was able to run all sample & repro apps, things seem to be working the same way as they were before the bump

.editorconfig
gpeal: This is neat. I actually wasn't aware that there was an IDE agnostic config like this. Will this get other people's IDEs to have the right settings even if the scheme is set to default?

CleanShot 2023-06-14 at 14 45 45@2x

.github/workflows/validate.yml
gpeal: Neat! Let's give it a shot 😄

.github/workflows/validate.yml
gpeal: I guess that makes sense. We can keep this as is but revert compileVersion in the build.gradle so we don't force others to upgrade.

lottie/build.gradle
gpeal: I had ambitions to add a bunch of errorprone checks to the core lottie library but never got around to adding them.

versions.properties
gpeal: I am a maintainer of Mavericks, too 😄 The updates all backwards compatible so we should be good here.

.editorconfig
mateuszkwiecinski: >Will this get other people's IDEs to have the right settings even if the scheme is set to default?

Yes, the `.editorconfig` preferences have precedence over IDE settings by default (but can be explicitly disabled). Intellij-based IDEs display additional information: 

image

>IDE agnostic config

I realized I already lied, partially. It's not 100% IDE agnostic - most of the properties start with `ij_`, which work only in IntelliJ based IDEs, but other properties like `indent_size` or `max_line_length` or `insert_final_newline` work in every popular IDE

build.gradle
mateuszkwiecinski: Java 8 target is marked as deprecated in Java 20. So when the project is compiled by Java 20 it prints an additional warning which I don't believe is actionable here (yet). This particular line disabled the warning. 
Given I reverted to compiling with Java 17 - the warning is gone and I removed this line.

.github/workflows/validate.yml
mateuszkwiecinski: Not having them in sync wouldn't be beneficial. The build is configured to be deterministic, so the version passed to `jvmToolchain` in `build.gradle` will be used to compile the app. If this specific version it's not installed, Gradle will attempt to download one. 
So if there was a _wrong_ version installed on the CI, the workflow would always have to download desired java version on each run, which would be suboptimal. Keeping CI config and build.gradle makes sure CI runs don't perform unnecessary actions.

mateuszkwiecinski: I kept Java 17 for now, but with the setup I added recently, it's safe to always use _latest_ java version and Gradle toolchains should make sure it's used in local builds, even if you don't have it installed manually. 

For example, my local setup has only Java 20 installed, and I compile the project with Java 17 downloaded by gradle:
image
so there is no inconvenience in managing Java version + the build is fully deterministic 👌

.github/workflows/validate.yml
mateuszkwiecinski: Thanks! Just to be clear, builds scans (the last colump) won't yet be published, the action is only capable of extracting them from the build output. 
I will configure build scans separately, I planned to do that when proposing changes to the CI pipeline.

lottie-compose/src/test/java/com/airbnb/lottie/compose/LottieAnimatableImplTest.kt
mateuszkwiecinski: This actually is tricky 😅 the _custom method_ is called `runTest` so if I import the one from coroutines it becomes ambiguous which `runTest` does a `runTest { }` invoke. 

As an alternative to fully qualified import I can propose:
1. Introducing an import alias `runTestRaw` or `runTestFromCoroutines`
2. Renaming custom `runTest` to `runTestWithClock`

.github/workflows/validate.yml
mateuszkwiecinski: And to reply to https://github.com/airbnb/lottie-android/pull/2316#discussion_r1230182074 within a single thread. latest AGP version require Java 17, but this is just the minimum supported version
 
>you need to use JDK 17 or higher.

I think it's always encouraged to build with latest Java version to leverage latest build improvements (i.e. initially only Java 17 and up had native M1 chips support (and I believe only some of the distributions backported it to lower java versions eventually))

.editorconfig
mateuszkwiecinski: Currently, only IDE will read `.editorconfig`, but I'd hope this project will eventually adapt for example [ktlint](https://github.com/pinterest/ktlint/) which would be another example of a tool that reads `.editorconfig` - [docs](https://pinterest.github.io/ktlint/0.49.1/rules/configuration-ktlint/)

gpeal: ktlint is great. I usually use it in combination with detekt-formatting (like what I did in [Mavericks](https://github.com/airbnb/mavericks/pull/191). If you'd like to take that on next, feel free!

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+
+[*.{kt,kts}]
+max_line_length = 140
+indent_size = 4
+ij_kotlin_allow_trailing_comma = true
+ij_kotlin_allow_trailing_comma_on_call_site = true
+
+[*.java]
+ij_java_class_count_to_use_import_on_demand = 3
+ij_java_names_count_to_use_import_on_demand = 3
```
gpeal, line 14: Is this the threshold to use * imports? If so, let's configure this to be 9999 or off somehow.

lottie-compose/src/test/java/com/airbnb/lottie/compose/LottieAnimatableImplTest.kt
gpeal: Let's go with number two since it's more explicit and more compatible with the new name

.editorconfig
mateuszkwiecinski: I'm not the biggest fan of star imports as well 😄 I now set it to the maximum value, both for Java and for Kotlin files. 

And yes, this was the threshold, I remembered the default value was 3, plus checked the committed [codestyle](https://github.com/airbnb/lottie-android/blob/master/.idea/codeStyles/Project.xml) file doesn't contain any other overrides.

snapshot-tests/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
mateuszkwiecinski, line 5: I'm not sure why it wasn't required before. The permission was required for a long, long time: https://github.com/aws-amplify/aws-sdk-android/blob/ce33403116f27c1524796ddf96fd429d4316d433/aws-android-sdk-s3/src/main/AndroidManifest.xml#L4 

It was never merged automatically, because the `libs.aws.android.sdk.s3` dependency is added only to the test-apk, which AFAIR can't have its own manifest? Or resources at all 🤔 I don't remember what precisely didn't work. 

Anyway, I hope this is the fix for the CI failure: 
```kotlin
java.lang.SecurityException: ConnectivityService: Neither user 10063 nor current process has android.permission.ACCESS_NETWORK_STATE.
at android.os.Parcel.readException(Parcel.java:1599)
at android.os.Parcel.readException(Parcel.java:1552)
at android.net.IConnectivityManager$Stub$Proxy.getActiveNetworkInfo(IConnectivityManager.java:974)
at android.net.ConnectivityManager.getActiveNetworkInfo(ConnectivityManager.java:646)
at com.amazonaws.mobileconnectors.s3.transferutility.TransferNetworkConnectionType.isConnected(TransferNetworkConnectionType.java:82)
at com.amazonaws.mobileconnectors.s3.transferutility.TransferRecord.checkPreferredNetworkAvailability(TransferRecord.java:375)
at com.amazonaws.mobileconnectors.s3.transferutility.TransferRecord.start(TransferRecord.java:188)
at com.amazonaws.mobileconnectors.s3.transferutility.TransferUtility.submitTransferJob(TransferUtility.java:982)
at com.amazonaws.mobileconnectors.s3.transferutility.TransferUtility.upload(TransferUtility.java:586)
at com.amazonaws.mobileconnectors.s3.transferutility.TransferUtility.upload(TransferUtility.java:537)
at com.amazonaws.mobileconnectors.s3.transferutility.TransferUtility.upload(TransferUtility.java:491)
at com.airbnb.lottie.snapshots.utils.HappoSnapshotter$uploadDeferred$2.invokeSuspend(HappoSnapshotter.kt:136)
at com.airbnb.lottie.snapshots.utils.HappoSnapshotter$uploadDeferred$2.invoke(HappoSnapshotter.kt)
at com.airbnb.lottie.snapshots.utils.HappoSnapshotter$uploadDeferred$2.invoke(HappoSnapshotter.kt)
```

gpeal: Odd, I would expect the manifest to get merged 🤷

gpeal (approved):

PR #2323, opened 2023-06-23
Support re-initialize

### LINKED ISSUE #2245: Re-calling `Lottie.initialize` leads to inconsistent state
[**Link to fork with a repro in the issue-repro module**](https://github.com/drewhamilton/lottie-android/tree/drew/double-initialize-bug)

**Describe the bug**
`Lottie.initialize` can be called multiple times without throwing an error, but subsequent calls do not change the `networkCache` and `networkFetcher` used internally in the `L` class.

**Steps To Reproduce**
Run `./gradlew :issue-repro:recordPaparazziDebug` on my branch linked above. Note that `heartUrl2` test fails.

The test class sets a `TemporaryFolder` as Lottie's network cache directory `@Before` each test. The `TemporaryFolder` is deleted after each test.

The first test, `heartUrl`, works as expected, generating a screenshot with a heart image loaded from the provided URL.

The second test, `heartUrl2`, fails. Reading down the stacktrace you can see that the cache directory is not a directory. This is because the directory no longer exists; it was deleted after the first test. Lottie is still holding and using a reference to the cache dir from the first `initialize` call, rather than the one from the second `initialize` call.

**Possible fixes**
One approach would be to throw an exception on subsequent calls to `initialize`. But this would make my use-case linked above—re-initializing Lottie for each Paparazzi JUnit test—impossible. To support my use-case, a `Lottie.reset` API could be added to clear previously-initialized settings. This could be `@VisibleForTesting` if you want to discourage its use in production code.

Another approach would be to decide that it is not an error to call `initialize` multiple times, and silently replace any old configuration with each call to `initialize`.

I'd love to submit a PR implementing whatever solution is preferred.

drewhamilton:
Closes #2245, see details there. This resets `L.networkFetcher` in `L.setFetcher`, and resets `L.networkCache` in `L.setCacheProvider`, ensuring that the public fetcher and cache provider are never out of sync with their respective internal implementations.

`LottieInitializeTest` verifies this by simulating two tests run with different `TemporaryFolder` instances. The new test fails if either of the two `L` properties are not reset.


lottie/src/main/java/com/airbnb/lottie/Lottie.java
```diff
@@ -20,7 +20,6 @@ public static void initialize(@NonNull final LottieConfig lottieConfig) {
     L.setCacheProvider(lottieConfig.cacheProvider);
     L.setTraceEnabled(lottieConfig.enableSystraceMarkers);
     L.setNetworkCacheEnabled(lottieConfig.enableNetworkCache);
-    L.setNetworkCacheEnabled(lottieConfig.enableNetworkCache);
```
drewhamilton, line 23: I assume this double-call was just a typo, but I can revert if it was intentional.

lottie/src/main/java/com/airbnb/lottie/L.java
```diff
@@ -75,10 +75,16 @@ private static LottieTrace getTrace() {
 
   public static void setFetcher(LottieNetworkFetcher customFetcher) {
     fetcher = customFetcher;
+    synchronized (NetworkCache.class) {
+      networkFetcher = null;
```
gpeal, line 79: What is the purpose of synchronizing a single line?

gpeal: In both cases here, it's probably best to only reset the value if the new value is different so that calling this is idempotent.

drewhamilton: I was trying to guard against a race between this and the `networkFetcher(Context)` function:
```java
NetworkFetcher local = networkFetcher;
if (local == null) {
  synchronized (NetworkFetcher.class) {
    local = networkFetcher;
    if (local == null) {
      networkFetcher = local = new NetworkFetcher(networkCache(context), fetcher != null ? fetcher : new DefaultLottieNetworkFetcher());
    }
    // What if networkFetcher gets set to null right now?
  }
}
return local;
```

In writing it out explicitly I realize it doesn't matter: in this case the next invocation of `networkFetcher(Context)` will generate a new instance as expected. I'll drop the new `synchronized` calls.

lottie/src/main/java/com/airbnb/lottie/L.java
```diff
@@ -74,15 +74,17 @@ private static LottieTrace getTrace() {
   }
 
   public static void setFetcher(LottieNetworkFetcher customFetcher) {
+    boolean changing = !fetcher.equals(customFetcher);
```
gpeal, line 77: I think one extra step of robustness would be to allow null here so that if you set it, you can also unset it.

gpeal: I also don't think you need to set changing. You can do
```
if (fetcher == null == customFetcher == null || (fetcher == null ? false : fetcher.equals(customFetcher)) {
  return;
}
```
Personally, I like guards that return early so you don't have to nest the rest of the business logic in an if.

gpeal:
I'm looking into the CI failure. For now, you can ignore it

lottie/src/main/java/com/airbnb/lottie/L.java
drewhamilton: Done!

gpeal (approved):

PR #2328, opened 2023-07-03
Fixes flickering when the same URL is set to the LottieAnimationView

### LINKED ISSUE #2326: Setting the same URL in LottieAnimationView multiple times results in flickering
**Checklist**

1. My animation doesn't use any [unsupported features](http://airbnb.io/lottie/#/supported-features). _Yes_
2. I know what part of my animation doesn't work. _Yes_
3. I have created a simplified version of my animation. _No, this bug can be reproduced with any animation_
4. I have attached the AEP file (as a zip file so it can be attached) that only has the part of the animation that doesn't work. _No, this bug can be reproduced with any animation_

**Describe the bug**
Setting the same URL multiple times in a LottieAnimationView results in the view flickering. This is because each time we call LottieAnimationView#setAnimationFromUrl() with the same URL, a new LottieTask is created and the current composition is cleared from the LottieAnimationView. The expected behavior is that the LottieAnimationView should detect that the same URL is set each time and that it does not need to reset and reload the same composition.

Would you be open for me to send a PR with a fix where the currently loaded URL is stored as part of the LottieAnimationView state and the call to setAnimationFromUrl() does nothing if the URL provided is the same as the currently loaded URL?

6.0

13

**Steps To Reproduce**

1. Create a UI with a single LottieAnimationView
2. Create a timer that will call [LottieAnimationView#setAnimationFromUrl()](https://github.com/airbnb/lottie-android/blob/1645311b20568448f55dbdc8b8a932485af4819d/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java#L534) periodically e.g. every 5 seconds, setting the same URL each time
3. Observe that the LottieAnimationView flickers, although the same composition is loaded each time.
4. The expected behavior is that the LottieAnimationView is not cleared when the same URL is set, because the composition is already loaded and does not need to be reloaded.

**Screenshots**
I have not created any screenshots yet, please let me know if that would be helpful.

chggr:
When the same URL is set in a LottieAnimationView multiple times, the view continuously clears and resets the composition from the cache resulting in flickering. This PR fixes the flickering by adding the currently loaded URL as part of the view state. If the same URL is set multiple times, the LottieAnimationView now knows not to reload the composition again and again and thus avoid flickering.

Fixes #2326

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -40,6 +40,7 @@ import java.lang.ref.WeakReference;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -90,7 +91,7 @@ import java.util.Set;
       if (targetView == null) {
         return;
       }
-      targetView.setComposition(result);
+      targetView.setCompositionInternal(result);
     }
   }
 
@@ -123,6 +124,7 @@ import java.util.Set;
 
   private final LottieDrawable lottieDrawable = new LottieDrawable();
   private String animationName;
+  private String animationUrl;
   private @RawRes int animationResId;
 
   /**
@@ -457,6 +459,7 @@ import java.util.Set;
   public void setAnimation(@RawRes final int rawRes) {
     this.animationResId = rawRes;
     animationName = null;
+    animationUrl = null;
     setCompositionTask(fromRawRes(rawRes));
   }
 
@@ -474,6 +477,7 @@ import java.util.Set;
   public void setAnimation(final String assetName) {
     this.animationName = assetName;
     animationResId = 0;
+    animationUrl = null;
     setCompositionTask(fromAssets(assetName));
   }
 
@@ -514,6 +518,7 @@ import java.util.Set;
    * Auto-closes the stream.
    */
   public void setAnimation(InputStream stream, @Nullable String cacheKey) {
+    animationUrl = null;
     setCompositionTask(LottieCompositionFactory.fromJsonInputStream(stream, cacheKey));
   }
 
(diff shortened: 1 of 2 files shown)


chggr:
Hi, I hope you are well. Could I please get some feedback on this PR? Many thanks!

gpeal (approved):

gpeal:
I think the same issue would happen with the same asset or raw res, right? I wonder if there is a more generic way that would handle all of these cases. Maybe something like getting the LottieComposition or null from LottieCompositionFactory and if it exists and is the same instance as the one set on the field in LottieAnimationView then noop. What do you think?

chggr:
Hi Gabriel, thanks for all your help so far! Would it make sense to proceed with this change to fix the issue for URL resources and revisit in the future to fix the issue more broadly for other resource types as well?  If yes, then we can merge this PR. 

I see there is a validation check failure, but I don't believe it is due to the changes in this PR. It seems more like an environment set up issue (i.e. EW_API_TOKEN is missing). What would be the best way to resolve this issue?

gpeal:
> Hi Gabriel, thanks for all your help so far! Would it make sense to proceed with this change to fix the issue for URL resources and revisit in the future to fix the issue more broadly for other resource types as well? If yes, then we can merge this PR.
> 
> I see there is a validation check failure, but I don't believe it is due to the changes in this PR. It seems more like an environment set up issue (i.e. EW_API_TOKEN is missing). What would be the best way to resolve this issue?

CI should be working again

gpeal:
> Hi Gabriel, thanks for all your help so far! Would it make sense to proceed with this change to fix the issue for URL resources and revisit in the future to fix the issue more broadly for other resource types as well? If yes, then we can merge this PR.

I think the final fix may look fairly different than this. I'm going to explore a solution that is more robust and based on the cache keys for LottieCompositionTask so I'd rather not land this in the meantime.

huynhdev24 (approved):

gpeal:
Closing in favor of https://github.com/airbnb/lottie-android/pull/2441

PR #2333, opened 2023-07-10
Fixes flickering when the same resource is set to the LottieAnimationView

### LINKED ISSUE #2326: Setting the same URL in LottieAnimationView multiple times results in flickering

chggr:
When the same resource (url, json, etc) is set in a LottieAnimationView multiple times, the view continuously clears and resets the composition resulting in flickering. This change fixes flickering by not eagerly clearing the LottieAnimationView, but instead keeping the old animation displayed until the new composition is available and can be rendered. The view is cleared only when there is a failure loading the new resource.

Fixes #2326

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -112,6 +112,8 @@ import java.util.Set;
 
       if (targetView.fallbackResource != 0) {
         targetView.setImageResource(targetView.fallbackResource);
+      } else {
+        targetView.clearComposition();
       }
       LottieListener<Throwable> l = targetView.failureListener == null ? DEFAULT_FAILURE_LISTENER : targetView.failureListener;
       l.onResult(result);
@@ -588,7 +590,6 @@ import java.util.Set;
 
   private void setCompositionTask(LottieTask<LottieComposition> compositionTask) {
     userActionsTaken.add(UserActionTaken.SET_ANIMATION);
-    clearComposition();
     cancelLoaderTask();
     this.compositionTask = compositionTask
         .addListener(loadedListener)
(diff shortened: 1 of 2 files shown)


gpeal (commented):
Unfortunately, I don't think it is safe to change this behavior. Lottie is used in far too many use cases and I'm sure that somebody somewhere is relying on this behavior. I think your previous approach was more on the right track. However, since it affects all composition types, it may not be super trivial to do well.

I can try to take a look at this at some point if you don't want to get too deep into the code and you may be better off working around this in your app for now. Up to you though.

chggr:
Thanks Gabriel for the feedback, I am closing this PR.

PR #2356, opened 2023-08-10
Allow configuring a default global value for async updates

gpeal:


UKMIITB:
@gpeal 
How do we set this in our app ?
Couldn't find documentation regarding this

gpeal:
@UKMIITB `Lottie.initialize(LottieConfig.Builder().….build())`

UKMIITB:
@gpeal 
Should this be defined in onCreate of application class ?

Also if we are building an SDK, then where should I ideally define this ?

gpeal:
@UKMIITB Yes, it should be somewhere around app startup.

This probably isn't recommended if you are creating an SDK because it will overwrite the caller's settings if they have set it. If you control LottieDrawable/LottieAnimationView/LottieAnimation composable in your SDK, you can enable async updates manually on each view.

PR #2364, opened 2023-08-18
Fix Potential NPE In NetworkCache.clearCache

undermark5:
In an app that I maintain, a crash was reported in Crashlytics indicating that there was a NPE in the Lottie library code.
Screenshot 2023-08-18 at 12 45 55 PM

It took me a bit to realize why the NPE was being caused by trying to read the length of a null array, but then I realized, the for loop calls `parentDir.listFiles()` again to iterate over rather than using the `files` array that was just checked the line above.

The changes are quite minor and should fix the problem of the NPE, that being said there may be other concerns that I'm not aware of.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/network/NetworkCache.java
+++ b/lottie/src/main/java/com/airbnb/lottie/network/NetworkCache.java
@@ -37,8 +37,8 @@ public class NetworkCache {
     File parentDir = parentDir();
     if (parentDir.exists()) {
       File[] files = parentDir.listFiles();
-      if (files != null && files.length > 0) {
-        for (File file : parentDir.listFiles()) {
+      if (files != null) {
+        for (File file : files) {
           file.delete();
         }
       }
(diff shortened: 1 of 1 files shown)


lottie/src/main/java/com/airbnb/lottie/network/NetworkCache.java
```diff
@@ -37,8 +37,8 @@ public void clear() {
     File parentDir = parentDir();
     if (parentDir.exists()) {
       File[] files = parentDir.listFiles();
-      if (files != null && files.length > 0) {
-        for (File file : parentDir.listFiles()) {
+      if (files != null) {
```
gpeal, line 40: The length check is a very minor optimization to not allocate an iterator when length as 0 so I'd recommend keeping it. other than that, lgtm

undermark5: Got it.

gpeal (approved):

Issue #1055, opened 2018-12-16
Add support for multiply blend mode


gpeal:

ThePixelBug:
Are any of these blend modes supported on android lottie? Maybe with the 3.0 release?

AE blend modes
normal;
multiply;
screen;
overlay;
darken;
lighten;
color-dodge;
color-burn;
hard-light;
soft-light;
difference;
exclusion;
hue;
saturation;
color;
luminosity;

gpeal:
@ThePixelBug Nope, not yet

jhughes696:
@gpeal  Are blend modes possible at all?  Multiply is the main one that would be a huge help.

gpeal:
https://github.com/airbnb/lottie-android/pull/2408 added support for all blend modes that are backwards compatible. Android Q added more but almost no apps have a minSdk of Q+ so it will be a while before support for those can be added.

PR #2408, opened 2023-10-30
Add support for layer blend mode "bm" for NORMAL, SCREEN, OVERLAY, DARKEN, LIGHTEN, and ADD

kudanai:
Adds support for "bm" (BlendMode) on layers. Possibly addresses #1055
Might need a little bit of a refactor. Relies on `Paint.setBlendMode()` available in android Q and up.

## Sample before & after
IMAGE 2023-10-30 12:28:18
Screenshot 2023-10-30 at 12 27 39


lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
```diff
@@ -120,6 +121,12 @@ public FillContent(final LottieDrawable lottieDrawable, BaseLayer layer, ShapeFi
       dropShadowAnimation.applyTo(paint);
     }
 
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+      final BlendModeCompat blendMode = layer.getBlendMode().toCoreBlendMode();
+      if(blendMode!=null)
```
gpeal, line 126:
```suggestion
      if (blendMode != null)
```

lottie/src/main/java/com/airbnb/lottie/model/content/LBlendMode.java
```diff
@@ -0,0 +1,78 @@
...
+  HUE,
+  SATURATION,
+  COLOR,
+  LUMINOSITY,
+  ADD,
+  HARD_MIX;
+
+  @Nullable
+  @RequiresApi(api = Build.VERSION_CODES.Q)
+  public BlendModeCompat toCoreBlendMode() {
```
gpeal, line 35:
```suggestion
  public BlendModeCompat toNativeBlendMode() {
```

lottie/src/main/java/com/airbnb/lottie/parser/LayerParser.java
```diff
@@ -49,7 +50,8 @@ private LayerParser() {
       "op", // 19
       "tm", // 20
       "cl", // 21
-      "hd" // 22
+      "hd", // 22
+      "bm" // 23
```
gpeal, line 54:
```suggestion
      "bm"  // 23
```

gpeal (commented):
Thanks for taking a stab at this!

I want to land this but in the past, I have been hesitant to launch feature that require new platforms until the usage of that platform is _very_ high. About 75% of Android phones are on Q and above but most apps aren't there with their minSdk yet.

The use case I'm concerned about is somebody landing an animation that looks correct during development but looks completely broken for end users with old phones.

The only other time I've launched something like this was [enableMergePathsForKitKatAndAbove](https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java#L400) in which it's opt-in and the API makes it very clear what the platform restrictions are.

Also, if you add animations to [this](https://github.com/airbnb/lottie-android/tree/master/snapshot-tests/src/main/assets/Tests) folder, they will automatically be picked up by snapshot tests.

kudanai:
Thank you for the feedback.

I understand what you mean by the version check. I could suggest rolling back a little and only supporting the blend modes that have lower minsdk requirements, and continue working on it from there. Does that seem like a more prudent approach?

Also I wasn't entirely sure about the structure of that enum, it feels overly verbose to me. However since `bm` can also appear for shapes, figured it might be useful to leave the enum itself in place.

lottie/src/main/java/com/airbnb/lottie/model/content/LBlendMode.java
```diff
@@ -0,0 +1,78 @@
...
+      case SATURATION:
+        return BlendModeCompat.SATURATION;
+      case COLOR:
+        return BlendModeCompat.COLOR;
+      case LUMINOSITY:
+        return BlendModeCompat.LUMINOSITY;
+      case ADD:
+        return BlendModeCompat.PLUS;
+      case HARD_MIX:
+        return null;  // todo: fix this
```
gpeal, line 72: Is there an explicit action item here or is it "wait for platform support"?

lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
```diff
@@ -120,6 +121,12 @@ public FillContent(final LottieDrawable lottieDrawable, BaseLayer layer, ShapeFi
       dropShadowAnimation.applyTo(paint);
     }
 
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+      final BlendModeCompat blendMode = layer.getBlendMode().toNativeBlendMode();
+      if (blendMode != null)
+        PaintCompat.setBlendMode(paint, blendMode);
```
gpeal, line 127: The entirety of setBlendMode wasn't added until API 29 so it's not just a matter of only supporting a subset of backwards-compatible ones, unfortunately.

lottie/src/main/java/com/airbnb/lottie/model/content/LBlendMode.java
kudanai: I think it would be "wait for platform support" at this time. There doesn't seem to be an equivalent to had_mix at this time if we're implementing it like this.

lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
kudanai: yeah I'm using [PaintCompat](https://developer.android.com/reference/androidx/core/graphics/PaintCompat) here which seems to offer up the ability to use [BlendModeCompat](https://developer.android.com/reference/androidx/core/graphics/BlendModeCompat). If the docs are up to spec, it should be able to give us anything that maps to the following in a backwards compatible way
   ```
  CLEAR,
  SRC,
  DST,
  SRC_OVER,
  DST_OVER,
  SRC_IN,
  DST_IN,
  SRC_OUT,
  DST_OUT,
  SRC_ATOP,
  DST_ATOP,
  XOR,
  PLUS,
  MODULATE,
  SCREEN,
  OVERLAY,
  DARKEN,
  LIGHTEN
  ```
  
  The implementation of `PaintCompat.setBlendMode()` looks like this
  ```java
public static boolean setBlendMode(@NonNull Paint paint, @Nullable BlendModeCompat blendMode) {
    if (Build.VERSION.SDK_INT >= 29) {
        Object blendModePlatform = blendMode != null
                ? BlendModeUtils.Api29Impl.obtainBlendModeFromCompat(blendMode) : null;
        Api29Impl.setBlendMode(paint, blendModePlatform);
        // All blend modes supported in Q
        return true;
    } else if (blendMode != null) {
        PorterDuff.Mode mode = obtainPorterDuffFromCompat(blendMode);
        paint.setXfermode(mode != null ? new PorterDuffXfermode(mode) : null);
        // If the BlendMode has an equivalent PorterDuff mode, return true,
        // otherwise return false
        return mode != null;
    } else {
        // Configuration of a null BlendMode falls back to the default which is supported in
        // all platform levels
        paint.setXfermode(null);
        return true;
    }
}
 ````

kudanai: Unfortunately that isn't a lot, but it's better than nothing imho,
It would give us (from lottie blend types) the following,

`NORMAL`, `SCREEN`, `OVERLAY`, `DARKEN`, `LIGHTEN`, and `ADD`

6/18 blend modes

lottie/src/main/java/com/airbnb/lottie/model/content/LBlendMode.java
gpeal: Let's make the comment reflect that then.

lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
gpeal: Unfortunately, it doesn't look like there is consistent behavior even for some of those.
1. If you look at the API 23 vs 31 diffs on the main PR page, there are several improved diffs for 31 but not for 23
2. If you look at the ones you added, if I'm reading this right, SCREEN should be bm2
31:



23:

kudanai:
Pushing new changes:

* Separate out the backwards compatible blend modes and ones that are only Q+
   * They've been moved out into a separate (unused for now) method
* Removed API version guard while actually applying the layer blend mode

kudanai:
The supported modes look consistent to me on 23 and 31 now

lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
```diff
@@ -120,6 +121,13 @@ public FillContent(final LottieDrawable lottieDrawable, BaseLayer layer, ShapeFi
       dropShadowAnimation.applyTo(paint);
     }
 
+    // Set layer paint blend mode
+    final BlendModeCompat blendMode = layer.getBlendMode().toNativeBlendMode();
+    if (blendMode != null) // PaintCompat.setBlendMode accepts null but no surprises please
```
gpeal, line 126: Is there a specific reason you don't want to pass in null here? Looking at the PaintCompat code, I think it would be fine.

lottie/src/main/java/com/airbnb/lottie/animation/content/FillContent.java
```diff
@@ -120,6 +121,13 @@ public FillContent(final LottieDrawable lottieDrawable, BaseLayer layer, ShapeFi
...
+    {
+      PaintCompat.setBlendMode(paint, blendMode);
```
gpeal, line 128: Can the blend mode change? If not, can we do it once in the constructor instead of on every frame?

lottie/src/main/java/com/airbnb/lottie/model/content/LBlendMode.java
```diff
@@ -0,0 +1,92 @@
...
+  COLOR,
+  LUMINOSITY,
+  ADD,
+  HARD_MIX;
+
+  @Nullable
+  public BlendModeCompat toNativeBlendMode() {
+    switch (this) {
+      case NORMAL:
+        return null; // BlendMode.CLEAR?
```
gpeal, line 37: Can you elaborate what this comment is for? I think null seems right for NORMAL, not CLEAR.

lottie/src/main/java/com/airbnb/lottie/model/content/LBlendMode.java
```diff
@@ -0,0 +1,92 @@
...
+        return BlendModeCompat.SCREEN;
+      case OVERLAY:
+        return BlendModeCompat.OVERLAY;
+      case DARKEN:
+        return BlendModeCompat.DARKEN;
+      case LIGHTEN:
+        return BlendModeCompat.LIGHTEN;
+      case ADD:
+        return BlendModeCompat.PLUS;
+      default:
```
gpeal, line 48: Instead of having `toNativeBlendModeQ` and landing it without any usages and this code commented, I would do the following:
```
switch (type) {
	case NORMAL:
		return null;
	case SCREEN:
		return BlendModeCompat.SCREEN;
	…
	// Blend modes below were not added to the platform until Q.
	// To prevent unexpected issues where animations look correct
	// during development but silently break for users with older devices
	// we won't support any of these until Q is widely used.
	case MULTIPLY:
	case COLOR_DODGE
	…
		return null;
```

If there was some complexity in the mapping, I could see the value in landing the code you write but because the mapping is trivial, I would prefer to not land unused code like that.

lottie/src/main/java/com/airbnb/lottie/model/content/LBlendMode.java
kudanai: I wasn't quite sure. Have gone back and confirmed that `null` is the way to go

gpeal (approved):
Thanks for doing this!

Issue #2409, opened 2023-10-31
Support load dotlottie for setAnimation(InputStream)


jhen0409:
https://github.com/airbnb/lottie-android/blob/4c805889968bfa2eba6c87cd99a093aa5261beb3/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java#L507-L518

Currently it looks like only support json, can we add support for zip?

I'm using this as workaround for now:

```kt
val stream = FileInputStream(File(filepath))
val result = LottieCompositionFactory.fromZipStreamSync(ZipInputStream(stream), filepath.hashCode().toString())
val composition = result.value ?: throw IllegalStateException("Unable to parse ${filepath}")
view.setComposition(composition)
```

Also, change from `fromZipStreamSync` to `fromZipStream` can be complex, maybe make `view.setCompositionTask` public will be better.

gpeal:
Sorry, I don't understand the question. Can you attach a zip file that isn't working correctly?

thevoiceless:
I believe the request is to expose a setter on `LottieAnimationView` that works with zip/`.lottie` sources. The existing `InputStream` setter assumes that the stream is backed by json.

jhen0409:
File is here: [test.lottie.zip](https://github.com/airbnb/lottie-android/files/13221729/test.lottie.zip)

After #1660 we can use `.lottie` files by check file extension to use zip input stream, but currently the `LottieAnimationView .setAnimation(InputStream)` cannot detect for that, it only expected the input is JSON. In my case I want to use local dotlottie files.

Do you think use API like `PushBackInputStream` to check bytes head is good idea, or just add a new API instead?

PR #2410, opened 2023-11-02
Add ability to load dotLottie via LottieAnimationView

thevoiceless:
Relevant issue: https://github.com/airbnb/lottie-android/issues/2409

Adds a `setAnimationFromDotLottie()` method to `LottieAnimationView` that accepts a `ZipInputStream`.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -41,6 +41,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.zip.ZipInputStream;
 
 /**
  * This view will load, deserialize, and display an After Effects animation exported with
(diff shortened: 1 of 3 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -556,6 +557,17 @@ public void setAnimationFromUrl(String url, @Nullable String cacheKey) {
     setCompositionTask(task);
   }
 
+  /**
+   * Load a <a href="https://dotlottie.io/">dotLottie</a> animation from a ZipInputStream.
+   * <p>
+   * Auto-closes the stream.
+   */
+  public void setAnimationFromDotLottie(ZipInputStream stream, @Nullable String cacheKey) {
```
gpeal, line 565:
```suggestion
  public void setAnimation(ZipInputStream stream, @Nullable String cacheKey) {
```

You can give this the same name. The method overloading specificity is enough here.

It's worth noting that dotLottie aren't the only zip files Lottie supports. You can zip any animations or animations + images so the naming/docs should not be dotLottie specific.

I was on a plane yesterday and actually drafted up a [similar PR](https://github.com/airbnb/lottie-android/pull/2411). I'm happy to land your contribution since you got it first, however.

thevoiceless: Makes sense! I don't have a strong opinion on which PR we land. Maybe yours, since I like your docs more 😄

gpeal: Okay, up to you. Feel free to update it or close it and then I'll land mine.

thevoiceless:
Closing in favor of https://github.com/airbnb/lottie-android/pull/2411

Issue #2262, opened 2023-03-11
Trim Path individually does not work, paths are animated simultaneously


AndrazP:
**Describe the bug**
Animating `Trim Paths` individually on multiple paths is not working, they are animated simultaneously.
I was only able to reproduce this issue on Android, it works correctly in Lottie AE Plugin and web(LottieFiles).

**What version of Lottie did you test this on?**
- 5.2.0
- 6.0.0

**What version of Android did you test this on?**
Android 12, 13

**Steps To Reproduce**

1. Create a layer with 3 line paths.
2. Animate `Trim Paths`start and set `Trim multiple shapes` to `individually`
3. Export and import to Android

**Screenshots**

Frame 30 [Web](https://app.lottiefiles.com/share/097c4468-48cc-47ab-b069-cbc37865c018)
Top path is completed before animation continues to next path.
![image](https://user-images.githubusercontent.com/1963904/224494342-70b1814d-6476-47fa-803c-38d80f0502a8.png)

Frame 30 Android
All paths are animated simultaneously.
![Screenshot_20230311_170529](https://user-images.githubusercontent.com/1963904/224495017-5f56f277-2081-4712-922e-fa3aea8c6124.png)

[AEP + Lottie JSON (trim_paths_test.zip)](https://github.com/airbnb/lottie-android/files/10948843/trim_paths_test.zip)

gpeal:
I spent a long time trying to improve this and I can't figure out a reasonable way to do this without significantly overhauling the rendering pipeline and adding a two-pass render phase. The challenge is that for this to work, you need to know the length of all of the paths that the trim path applies to before you draw the first one.

You can see my attempt [here](https://github.com/airbnb/lottie-android/pull/2417) but it will fail in cases where different shape groups have different transforms.

It would be great to support this, eventually, but I think you'll need to manually adjust trim paths on each layer to achieve this on Android.

PR #2417, opened 2023-11-15
Improve support for trim path individually

gpeal:
Previously, trim paths individually only worked for shapes within one content group. However, they will now work when spanning multiple groups.

Fixes https://github.com/airbnb/lottie-android/issues/2262


gpeal:
https://github.com/airbnb/lottie-android/issues/2262#issuecomment-1975380485

PR #2425, opened 2023-12-13
Added Compose Multiplatform support

alexzhirkevich:
Added support for Compose Multiplatform on desktop, iOS and web. It wraps skottie on these targets and shares playback logic with Android.

Note: skia is an implementation detail of compose and its apis are not considered stable. It will require to switch renderer if JetBrains change their engine. But it won't happen in near future.

Also, if these changes are not appropriate for this repo, i have it as a [separate library](https://github.com/alexzhirkevich/compottie). But i assume people prefer having official support


Blincwear88 (approved):

gpeal:
@alexzhirkevich Thank you for the work you have put in here. KMP is very exciting and what you've done here is both impressive and interesting. However, I spent some time thinking about this and I don't think that airbnb/lottie-android is the right home for this repo and recommend publishing it on your own as you have already done.

The core part of this repo is the Android Lottie renderer and the Compose APIs are a rather thin layer on top. Bringing in multiplatform means that people will come here to report bugs in skottie or on other platforms which neither me nor anybody at Airbnb is equipped to support. Skottie also has a fairly different feature set than Android in terms of rendering capabilities as well as APIs (like dynamic properties) so cleanly wrapping them in a universal API isn't trivial without introducing confusion for developers. You're welcome to continue using airbnb/lottie-android as a dependency in your library if you'd like though.

alexzhirkevich:
No problem. Thanks for your work on this project! 
I'll continue to work on it as a separate lib. It would be nice if you add it to the list of unofficial tools in lottie repo. I'm sure many people will find it useful

gpeal:
Sure thing, feel free to put up a PR to https://github.com/airbnb/lottie

PR #2427, opened 2023-12-13
GradientColorKeyframeAnimation does not handle progress outside [0,1]

### LINKED ISSUE #2426: GradientColorKeyframeAnimation does not handle progress outside [0,1]
**Describe the bug**
According to documentation of `BaseKeyframeAnimation.getValue` progress outside of `[0,1]` should be supported as well:
> keyframeProgress will be [0, 1] unless the interpolator has overshoot in which case, this should be able to handle values outside of that range.

However, `GradientColorKeyframeAnimation.getValue`/`GradientColor.lerp` do not handle this case.

6.2.0

Android 13

**Steps To Reproduce**
I can not attach a simplified version of our Lottie. But I am gonna prepare a PR in a second.

**Screenshots**
|6.2.0|Expected|
|---|---|
|image|image|

brudaswen:
- Closes #2426
- Improve performance by running GammaEvaluator once


kihaki:
Thank you for addressing this, this is an important issue for us 🙏🏻

lottie/src/main/java/com/airbnb/lottie/model/content/GradientColor.java
```diff
@@ -28,6 +28,19 @@ public int getSize() {
   }
 
   public void lerp(GradientColor gc1, GradientColor gc2, float progress) {
+    // Fast return in case start and end is the same
+    // or if progress is at start/end or out of [0,1] bounds
+    if (gc1 == gc2) {
```
gpeal, line 33: This is java so if you want to use your equals method, you have to use .equals

brudaswen: Oh my, so long ago... 🙈 
```suggestion
    if (gc1.equals(gc2)) {
```

gpeal (commented):
Thanks! Would it be possible to create a simplified animation for testing? You can put it [here](https://github.com/airbnb/lottie-android/tree/master/snapshot-tests/src/main/assets/Tests) and it'll automatically get added to the snapshot regression tests so it doesn't break in the future.

brudaswen:
> Thanks! Would it be possible to create a simplified animation for testing? You can put it [here](https://github.com/airbnb/lottie-android/tree/master/snapshot-tests/src/main/assets/Tests) and it'll automatically get added to the snapshot regression tests so it doesn't break in the future.

I stripped down our ultra large file to a minimum. I hope this works.

gpeal (approved):

Issue #2193, opened 2022-11-30
Support rendering using a Painter


dandc87:
**Is your feature request related to a problem? Please describe.**
I'd like to use a Lottie animation as a placeholder for images loaded via Coil, which only directly supports Painters

**Describe the solution you'd like**
An implementation of `Painter` which has the same constructor parameters as the `LottieAnimation` composable. I believe that since most of the logic is inside of the Canvas `onDraw`, it would be straight-forward to move into a Painter's `onDraw`.

**Describe alternatives you've considered**
Coil also allows for manual state management, which you could use to add and remove `LottieAnimation` to the composition. We currently do this, but it's not exactly intuitive.

**Additional context**
Coil similarly has their API support both a Painter (`rememberAsyncImagePainter`) and a composable (`AsyncImage`), both of which are handy in different situations.

notAnElephant:
how do I use it though? Is there an example I could use?

gpeal:
> how do I use it though? Is there an example I could use?

Call [rememberLottiePainter](https://github.com/airbnb/lottie-android/pull/2442/files?w=1#diff-f606c8f514c5c7d1b98ac41c261977bd56049c005015cc47306b15ba34a72aafR28) and then use it anywhere you would use a painter.

PR #2442, opened 2023-12-30
[Compose] Add LottiePainter

gpeal:
This allows Lottie to be used with anything that accepts a `Painter` (like `Image`).

Fixes https://github.com/airbnb/lottie-android/issues/2193


lottie-compose/src/main/java/com/airbnb/lottie/compose/LottiePainter.kt
```diff
@@ -0,0 +1,132 @@
+package com.airbnb.lottie.compose
```
Trav5599, line 1: Ok

Trav5599 (approved):

PR #2451, opened 2024-01-22
[a11y] Check for reduced motion frame names

bassettsj:
Brings parity to the android for the changes in iOS airbnb/lottie-ios/pull/2110

This PR adds support for respecting the system "reduction motion" option.

### Animations Enabled

https://github.com/airbnb/lottie-android/assets/1218420/d29b62f6-db07-49f9-a7de-221949eef646

### Animations Disabled

https://github.com/airbnb/lottie-android/assets/1218420/2f346a78-d637-4c4a-8f60-8aa4e69e140d


lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -794,14 +795,52 @@ public void playAnimation() {
       }
     }
     if (!animationsEnabled()) {
-      setFrame((int) (getSpeed() < 0 ? getMinFrame() : getMaxFrame()));
+      Marker markerForAnimationsDisabled = getMarkerForAnimationsDisabled();
+      if (markerForAnimationsDisabled != null) {
+        setFrame((int) markerForAnimationsDisabled.startFrame);
```
calda, line 800: On iOS we also support reduced motion markers with both a start frame and an end frame. If the marker provides an end frame, we play the animation from the start / end of the reduced motion marker as opposed to just only displaying the first frame.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -794,14 +795,52 @@ public void playAnimation() {
       }
     }
     if (!animationsEnabled()) {
-      setFrame((int) (getSpeed() < 0 ? getMinFrame() : getMaxFrame()));
+      Marker markerForAnimationsDisabled = getMarkerForAnimationsDisabled();
+      if (markerForAnimationsDisabled != null) {
```
calda, line 799: What should determine if "reduced motion" mode is enabled? Reading the code it seems like the reduced motion marker is always used if present, as opposed to only using it when the app / OS "reduced motion" toggle is enabled.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -794,14 +795,52 @@ public void playAnimation() {
...
+   * This method is used to get the marker for animations when system animations are disabled.
+   * It iterates over the list of allowed reduced motion markers and returns the first non-null marker it finds.
+   * If no non-null marker is found, it returns null.
+   *
+   * @return The first non-null marker from the list of allowed reduced motion markers, or null if no such marker is found.
+   */
+  public Marker getMarkerForAnimationsDisabled() {
+    Marker marker = null;
+    for (String markerName : ALLOWED_REDUCED_MOTION_MARKERS) {
+      marker = composition.getMarker(markerName);
```
calda, line 836: The comment above says "Supported marker names are case insensitive", so I'd expect marker names like "REDUCED_MOTION" to be supported. Is this `getMarker` check case-sensitive or case-insensitive?

gpeal: Under the hood, it's case insentivie
https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/model/Marker.java#L29

bassettsj: I think it is case insensitive, so we would be okay here.

calda: Great! That's convenient

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
bassettsj: @calda  Thanks for the quick reply, Is there a way to play from the start to end frame? Sorry for not knowing from looking at this code.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
gpeal: It defaults to [this](https://github.com/airbnb/lottie-android/blob/f41f03a13258987ea89e41a17e6df0d3a84adac2/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java#L255) which can be set if battery saver is on or if you turn off animations in developer settings. It's not an a11y setting on Android afaik though.

calda: Oh I see, I missed that this is gated by a `animationsEnabled()` check, which can be configured via `setSystemAnimationsAreEnabled`.

I think we'd probably want the "is reduced motion enabled" check to be configured separately from the "are system animations enabled" check.

gpeal: It doesn't feel quite right to take a specific a11y setting on iOS and use the same name on Android when it maps to different things.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
calda: @gpeal should be able to help :)

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
bassettsj: @gpeal  In the a11y community, "reduced motion" is the more common term even if it comes from Apple. I think the same is applied to lottie-web as well, so designers can produce one asset to share across platform.

bassettsj: > It's not an a11y setting on Android afaik though.
It is available under accessibility then Color and Motion
Color and Motions settings screen

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
bassettsj: @gpeal  How would I play the animation from start to end here per @calda 's comment?

calda (approved):
Thanks for working on this! Looks great from an iOS perspective, especially after adding support for markers with an end frame

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
gpeal: I think that's reasonable. It's worth noting that this _would_ affect non a11y use cases like battery saver and developer settings but this behavior may still be better if the designer was explicit about it.

Could you:
1. Double check that the utils function I linked to returns 0 for animation scale when this a11y setting is enabled
2. Add documentation for `setSystemAnimationsAreEnabled` in case anybody else calls it so they can learn about this without reading the source?

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -794,14 +795,52 @@ public void playAnimation() {
...
 
+  /**
+   * The marker to use if "reduced motion" is enabled.
+   * Supported marker names are case insensitive, and include:
+   *   - reduced motion
+   *   - reducedMotion
+   *   - reduced_motion
+   *   - reduced-motion
+   */
+  private static final List<String> ALLOWED_REDUCED_MOTION_MARKERS = Arrays.asList(
```
gpeal, line 819: Could you move this to the top of LottieDrawable (after invalidateSelfOnMainThread)?

gpeal (commented):
Thanks for doing this. I also raised this with the lottie spec team and they created https://github.com/lottie/lottie-spec/issues/7 for visibility and awareness for other platforms.

gpeal:
If you want to create a ReducedMotionTestCase, that would be great 😄 
https://github.com/airbnb/lottie-android/tree/master/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/tests

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
bassettsj: Checked Accessibility setting in the utils function, it returns false when running in the debugger.

Screenshot 2024-01-23 at 1 18 54 PM

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
bassettsj: Moved. thanks for the help.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
bassettsj: > Add documentation for setSystemAnimationsAreEnabled in case anybody else calls it so they can learn about this without reading the source?

I added the document comment, thanks for the suggestion

bassettsj:
@gpeal  I think I added the test case that might work. Wasn't able to run locally, but we will see with the workflow here.

gpeal:
Could you apply this patch to fix the test:

```
diff --git a/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/SnapshotTestCaseContext.kt b/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/SnapshotTestCaseContext.kt
index c5693a82..92df041d 100644
--- a/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/SnapshotTestCaseContext.kt
+++ b/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/SnapshotTestCaseContext.kt
@@ -56,7 +56,7 @@ suspend fun SnapshotTestCaseContext.withDrawable(
     assetName: String,
     snapshotName: String,
     snapshotVariant: String,
-    callback: (LottieDrawable) -> Unit,
+    callback: suspend (LottieDrawable) -> Unit,
 ) {
     val result = LottieCompositionFactory.fromAssetSync(context, assetName)
     val composition = result.value ?: throw IllegalArgumentException("Unable to parse $assetName.", result.exception)
@@ -254,4 +254,4 @@ private suspend fun View.awaitFrame() {
             cont.resume(Unit)
         }
     }
-}
\ No newline at end of file
+}
diff --git a/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/tests/DisabledAnimationsTestCase.kt b/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/tests/DisabledAnimationsTestCase.kt
index f535ae8e..b669635b 100644
--- a/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/tests/DisabledAnimationsTestCase.kt
+++ b/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/tests/DisabledAnimationsTestCase.kt
@@ -3,17 +3,23 @@ package com.airbnb.lottie.snapshots.tests
 import com.airbnb.lottie.snapshots.SnapshotTestCase
 import com.airbnb.lottie.snapshots.SnapshotTestCaseContext
 import com.airbnb.lottie.snapshots.withDrawable
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 
 class DisabledAnimationsTestCase : SnapshotTestCase {
     override suspend fun SnapshotTestCaseContext.run() {
         withDrawable("Tests/ReducedMotion.json", "System Animations", "Disabled") { drawable ->
-            drawable.setSystemAnimationsAreEnabled(false)
-            drawable.playAnimation()
+            withContext(Dispatchers.Main) {
+                drawable.setSystemAnimationsAreEnabled(false)
+                drawable.playAnimation()
+            }
         }
 
         withDrawable("Tests/ReducedMotion.json", "System Animations", "Enabled") { drawable ->
-            drawable.setSystemAnimationsAreEnabled(false)
-            drawable.playAnimation()
+            withContext(Dispatchers.Main) {
+                drawable.setSystemAnimationsAreEnabled(false)
+                drawable.playAnimation()
+            }
         }
     }
 }
```

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -794,14 +810,38 @@ public void playAnimation() {
...
 
+
+  /**
+   * This method is used to get the marker for animations when system animations are disabled.
+   * It iterates over the list of allowed reduced motion markers and returns the first non-null marker it finds.
+   * If no non-null marker is found, it returns null.
+   *
+   * @return The first non-null marker from the list of allowed reduced motion markers, or null if no such marker is found.
+   */
+  public Marker getMarkerForAnimationsDisabled() {
```
gpeal, line 834:
```suggestion
  private Marker getMarkerForAnimationsDisabled() {
```

gpeal (commented):
Looks good! Just one small last comment

gpeal (approved):

PR #2454, opened 2024-01-29
Add support for gzipped and tgs files from network

LinDevHard:
In addition to [#2435](https://github.com/airbnb/lottie-android/pull/2435), added support for loading gziped and tgs files from the network, similar to zip files.


lottie/src/main/java/com/airbnb/lottie/network/FileExtension.java
```diff
@@ -8,7 +8,8 @@
 @RestrictTo(RestrictTo.Scope.LIBRARY)
 public enum FileExtension {
   JSON(".json"),
-  ZIP(".zip");
+  ZIP(".zip"),
+  GZIP(".gz");
```
gpeal, line 12: Will you add tgs here?

gpeal (approved):

gpeal (approved):

PR #2497, opened 2024-04-25
Fix blinking issue when loop the animation in minFrame and maxFrame

yunlinos:
I encount blinking issue with a company internal resource.

Problem reproduction steps：
1. setMinAndMaxFrame(m, n) and repeatCount = INFINITE
2. it will notifyUpdate when animation ended, but the frame has not been reset to minFrame at this time. 
2. so it will draw n+1 frame.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/LottieValueAnimator.java
@@ -105,18 +105,14 @@ public class LottieValueAnimator extends BaseLottieAnimator implements Choreogra
 
     lastFrameTimeNs = frameTimeNanos;
 
-    if (!useCompositionFrameRate || frameRaw != previousFrameRaw) {
-      notifyUpdate();
-    }
     if (ended) {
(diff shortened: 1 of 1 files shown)


bunduqia:
I added 8 Lottie and when downloading the application they load very slowly because of this I was denied application hosting. Then I added a separate cache with Lottie to speed up the method, the method does not provide accelerated loading, I added loading from the cache to RAM, the method did not provide acceleration. I added Lottie's RAM preload and instant output for faster loading. Lottie sizes: 148, 43, 81, 66, 256, 17, 42, 180 KB and they can't load at the same time, the smaller ones appear first and the larger ones appear last, it feels like working with Lottie is very slow. system, it reminded me of i386 computers from 1998 that took a long time to boot. Please do something about this. In the meantime, I’m switching again to GIF, where files of this size or larger are loaded instantly without a brake or sequential appearance.

gpeal (commented):
Thanks for doing this. Sorry for not reviewing this for so long. For some reason, the CI checks didn't trigger. Can you push an empty commit or rebase/merge master and push to trigger CI?

gpeal:
@yunlinos Do you want to update this?

gpeal (approved):

PR #2510, opened 2024-06-04
Add support for Luma nad Inverted Luma matte mode

alexzhirkevich:
Hi. Added support for Luma and Inverted Luma matte mode

cc @gpeal

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
@@ -5,6 +5,8 @@ import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
 import android.graphics.Color;
+import android.graphics.ColorFilter;
+import android.graphics.ColorMatrixColorFilter;
 import android.graphics.Matrix;
 import android.graphics.Paint;
 import android.graphics.Path;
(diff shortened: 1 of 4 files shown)


lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
```diff
@@ -67,6 +69,15 @@ protected float[] initialValue() {
 
   private static final float INV_SQRT_2 = (float) (Math.sqrt(2) / 2.0);
 
+  private static final ColorFilter lumaColorFilter = new ColorMatrixColorFilter(
```
gpeal, line 72: Thanks for looking into this. What is the source for the values in this matrix?

bunduqia:
I added 8 Lottie and when downloading the application they load very slowly because of this I was denied application hosting. Then I added a separate cache with Lottie to speed up the method, the method does not provide accelerated loading, I added loading from the cache to RAM, the method did not provide acceleration. I added Lottie's RAM preload and instant output for faster loading. Lottie sizes: 148, 43, 81, 66, 256, 17, 42, 180 KB and they can't load at the same time, the smaller ones appear first and the larger ones appear last, it feels like working with Lottie is very slow. system, it reminded me of i386 computers from 1998 that took a long time to boot. Please do something about this. In the meantime, I’m switching again to GIF, where files of this size or larger are loaded instantly without a brake or sequential appearance.

lottie/src/main/java/com/airbnb/lottie/utils/Utils.java
gpeal: @alexzhirkevich You resolved this comment but could you provide the source for thsi?

alexzhirkevich: It doesn't work well anyway as you can remember. I'll close this one until a better approach appears

PR #2512, opened 2024-06-17
Add generalized Lottie Feature Flags API

jbeta51:
Currently there are individual methods for enabling and disabling Merge Paths in Lottie. This PR aims to generalize these functions as we consider adding more features guarded behind opt-in flags.

See https://github.com/jbeta51/lottie-android/pull/2 for draft comments.


gpeal (approved):

gpeal (commented):
Thanks for getting started with this!

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -403,14 +403,22 @@ public void setUseCompositionFrameRate(boolean useCompositionFrameRate) {
...
 
   /**
    * Returns whether merge paths are enabled for KitKat and above.
    */
   public boolean isMergePathsEnabledForKitKatAndAbove() {
-    return lottieDrawable.isMergePathsEnabledForKitKatAndAbove();
+    return lottieDrawable.isFeatureFlagEnabled(LottieFeatureFlags.FeatureFlag.MergePathsApi19);
+  }
+
+  public void enableFeatureFlag(LottieFeatureFlags.FeatureFlag flag, boolean enable) {
```
gpeal, line 416: Could you add docs here since this is a new public API?

I would make it extra explicit that people validate that opted in features look acceptable across all supported API levels.

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -285,34 +285,44 @@ public boolean hasMatte() {
...
 
+  @Deprecated
   public boolean enableMergePathsForKitKatAndAbove() {
-    return enableMergePaths;
+    return lottieFeatureFlags.isFlagEnabled(LottieFeatureFlags.FeatureFlag.MergePathsApi19);
   }
 
   /**
    * Enable this to get merge path support for devices running KitKat (19) and above.
+   * Deprecated: Use enableFeatureFlag(LottieFeatureFlags.FeatureFlag.MergePathsApi19, enable)
```
gpeal, line 295: You can use `@deprecated` and `@link` for proper javadoc syntax.

lottie/src/main/java/com/airbnb/lottie/LottieFeatureFlags.java
```diff
@@ -0,0 +1,48 @@
+package com.airbnb.lottie;
+
+import android.annotation.SuppressLint;
+import android.os.Build;
+
+import com.airbnb.lottie.utils.Logger;
+
+import java.util.HashSet;
+
+public class LottieFeatureFlags {
```
gpeal, line 10: Let's move `LottieFeatureFlags` to a package that's package-private so we don't expose this class in the public API and make `FeatureFlag` a top-level public enum `LottieFeatureFlag`

gpeal (approved):
Looks good! Would you like to add this API to lottie-compose as well or would you like me to?

gpeal (approved):

jbeta51:
> Looks good! Would you like to add this API to lottie-compose as well or would you like me to?

I could add it, but I also thought that was covered by adding it to LottiePainter.kt and LottieAnimation.kt, so I'd need to be pointed in the right direction. We can followup later or in another PR.

gpeal:
> > Looks good! Would you like to add this API to lottie-compose as well or would you like me to?
> 
> I could add it, but I also thought that was covered by adding it to LottiePainter.kt and LottieAnimation.kt, so I'd need to be pointed in the right direction. We can followup later or in another PR.

I'll add support for Compose in a bit
https://github.com/airbnb/lottie-android/issues/2515

PR #2514, opened 2024-06-28
LottieComposition: Add getUnscaledHeight & getUnscaledWidth functions.

vanniktech:
Use case: I just want to know the unscaled original height / width of the LottieAnimation. I want to do my own scaling by calculating for instance given an aspect ratio or weighted settings, how big/small I can stretch the animation.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/parser/LottieCompositionMoshiParser.java
+++ b/lottie/src/main/java/com/airbnb/lottie/parser/LottieCompositionMoshiParser.java
@@ -44,8 +44,8 @@ public class LottieCompositionMoshiParser {
     float frameRate = 0f;
     final LongSparseArray<Layer> layerMap = new LongSparseArray<>();
     final List<Layer> layers = new ArrayList<>();
-    int width = 0;
-    int height = 0;
+    int unscaledWidth = 0;
+    int unscaledHeight = 0;
     Map<String, List<Layer>> precomps = new HashMap<>();
     Map<String, LottieImageAsset> images = new HashMap<>();
     Map<String, Font> fonts = new HashMap<>();
@@ -57,10 +57,10 @@ public class LottieCompositionMoshiParser {
     while (reader.hasNext()) {
       switch (reader.selectName(NAMES)) {
         case 0:
-          width = reader.nextInt();
+          unscaledWidth = reader.nextInt();
           break;
         case 1:
-          height = reader.nextInt();
+          unscaledHeight = reader.nextInt();
           break;
         case 2:
           startFrame = (float) reader.nextDouble();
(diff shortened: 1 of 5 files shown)


gpeal (commented):
Can you divide the returned width/height by `Resources.getSystem().getDisplayMetrics().density` instead? That's all the scaling is doing.

vanniktech:
I also thought about that. But then I'd kind of rely on the implementation that's inside Lottie. If that changes for some reason, I'm screwed. iOS also offers to get the original size:


Screenshot 2024-06-29 at 14 01 07

Also after dividing, I'd need to round it (I want the integer value).

lottie/src/main/java/com/airbnb/lottie/parser/LottieCompositionMoshiParser.java
```diff
@@ -107,7 +107,7 @@ public static LottieComposition parse(JsonReader reader) throws IOException {
     Rect bounds = new Rect(0, 0, scaledWidth, scaledHeight);
 
     composition.init(bounds, startFrame, endFrame, frameRate, layers, layerMap, precomps,
-        images, Utils.dpScale(), characters, fonts, markers);
```
vanniktech, line 110: Note that I do pass the unscaled width & height here.

gpeal:
> Can you divide the returned width/height by `Resources.getSystem().getDisplayMetrics().density` instead? That's all the scaling is doing.

The implementation of that definitely won't change. I'd like to keep the API surface area of Lottie as small as possible. If the difference here really is dividing by that number, I would prefer not to add it to the core library.

vanniktech:
The difference is by dividing that number and needing to round it which may yield 399 instead of the original 400 depending on the scale and the rounding mode. It's really not a lot that is added to the API surface, has parity with the iOS implementation and yields something that is part of the format.

lottie/src/main/java/com/airbnb/lottie/parser/LottieCompositionMoshiParser.java
gpeal: Can you make the names very clearly named `unscaledWidth` and `unscaledHeight`?

gpeal (commented):
I see what you mean with regards to potentially being off by one due to multiplying a rounded value.

Depending on your use case, you could create a JsonReader for the same animation file and read the `h` and `h` top-level properties yourself.

We can add this to the core library but let's make the API naming very explicit.

vanniktech:
Do you also want me to prefix the exposed API with `unscaled`?

gpeal:
> Do you also want me to prefix the exposed API with `unscaled`?

Yes, please!

vanniktech:
Should be good to go now.

Issue #485, opened 2017-10-06
Add Range Selector support for Text animators


luciancioroga:
This would be a nice to have feature, and would complete the Text support.

PR #2518, opened 2024-07-19
Add support for text range selectors

allenchen1154:
This adds support for the [text range selector](https://lottiefiles.github.io/lottie-docs/text/#text-range-selector) type, which allows a [text style](https://lottiefiles.github.io/lottie-docs/text/#text-style) to be applied to a certain range of the text, defined by a start, end, and offset.

The text range selector implementation currently has the following limitations:
- Only text layers drawn using a Font are supported. Adding support for the glyph draw path may be possible, but I don't have a sample Lottie file.
- Only one text range is currently supported per text layer, as the parser [drops all other entries in the array](https://github.com/airbnb/lottie-android/blob/c4cb2254eca3c70199f1de5e39e3872c8c42e473/lottie/src/main/java/com/airbnb/lottie/parser/LayerParser.java#L194-L196).
- Only index-based ranges are supported - percentage-based ranges are also allowed in the spec.
- Only ranges based on characters are supported. The [spec](https://lottiefiles.github.io/lottie-docs/constants/#text-based) allows characters, characters excluding spaces, words, and lines.
- Other options like easing (allows styling characters that are partially inside the range), randomize, and [shape](https://lottiefiles.github.io/lottie-docs/constants/#text-shape) of the range are not currently supported.

This also adds support for the opacity as an animatable text property which is applied multiplicatively with the transform opacity and the parent's alpha.

Partially addresses https://github.com/airbnb/lottie-android/issues/485.

https://github.com/user-attachments/assets/bcfad060-482d-48d9-a578-297c4f143ba9

https://github.com/user-attachments/assets/211dc574-5ea1-4fa3-9f78-f87ee104ce85


lottie/src/main/java/com/airbnb/lottie/model/layer/TextLayer.java
```diff
@@ -459,7 +535,8 @@ private void drawGlyph(Path path, Paint paint, Canvas canvas) {
     canvas.drawPath(path, paint);
   }
 
-  private void drawCharacterFromFont(String character, DocumentData documentData, Canvas canvas) {
+  private void drawCharacterFromFont(String character, DocumentData documentData, Canvas canvas, int indexInDocument, int parentAlpha) {
+    configurePaint(documentData, parentAlpha, indexInDocument);
```
allenchen1154, line 539: This means we need to configure the paint on every character. I haven't looked into the performance implication or further optimization.

gpeal (commented):
Thanks for doing this! Try out this project/json file to see if you can get glyphs working. Hopefully it's a small lift
[Glyphs.zip](https://github.com/user-attachments/files/16328013/Glyphs.zip)

allenchen1154:
Added support for glyph-based text layers and percentage-based ranges. The sample looks a bit off because easing isn't supported - I believe the opacity should ideally fade out as each character leaves the range, but I couldn't find a clear spec on how to implement it. I would leave that for a follow-up.

https://github.com/user-attachments/assets/69c777dc-7a5c-48a4-aa1f-1e41eef91d86

gpeal (approved):
This is great! Do you know how much extra work it would take to support easing/percentages? We can land this as-is but it would be great to add if possible!

allenchen1154:
@gpeal I'd like to, but I don't have a reference for how the easing parameters should affect the opacity and other style properties. I was referring [these LottieFiles docs](https://lottiefiles.github.io/lottie-docs/text/#text-range-selector) but they don't provide any additional information.

gpeal:
> @gpeal I'd like to, but I don't have a reference for how the easing parameters should affect the opacity and other style properties. I was referring [these LottieFiles docs](https://lottiefiles.github.io/lottie-docs/text/#text-range-selector) but they don't provide any additional information.

I created https://github.com/airbnb/lottie-android/issues/2522 to track that as a potential follow up. I'll take a look if I can find some time for it

PR #2519, opened 2024-07-23
Apply blend modes on layer level and add Multiply blend mode

geomaster:
This commit improves blend mode support in lottie-android in two ways:
* Applying blend modes on layer-level, instead of fill level
* Adding support for the Multiply blend mode

## Applying blend modes on layer level

The Lottie format defines blend modes as attributes on a layer. However, lottie-android is presently applying the layer blend modes on a solid color fill only. Notably, this causes any stroked or gradient-filled shapes or image layers to blend incorrectly, such as in this file:

[stroke-blending-test.json](https://github.com/user-attachments/files/16346206/stroke-blending-test.json)

(The file contains a filled + stroked shape that renders as a pink square on other platforms, but renders with a visible stroke on lottie-android since its blend mode is applied only on the fill.)

Instead, we move this decision to `BaseLayer` by analogy to transparent layer handling, which is closer to how the format specifies the property and fixes these cases.

## Multiply support

`BlendModeCompat` is designed to resolve to either a `BlendMode` (added in Android Q, supporting most modern blend modes) or `PorterDuff.Mode` (always available, but a smaller choice of modes as it is mostly focused on alpha compositing).

We use `BlendModeCompat` to support Lottie layer blend modes (`bm` key) to ensure compatibility on all platforms. For consistency, we don't  support values which don't have a `PorterDuff.Mode` equivalent.

Our support for Lottie blend modes did not include Multiply due to a slightly different behavior between the `PorterDuff.MULTIPLY` (exposed as `BlendModeCompat.MODULATE`) and `BlendModeCompat.MULTIPLY` variants. Namely, the formula used for `PorterDuff.MODULATE`, combined with alpha-premultiplication done by Skia, means that a layer with an alpha < 1.0 and multiply blend mode will also darken the destination:

Incorrect-Blend

(Multiply-blended layers with < 1.0 alpha on the left, Screen-blended layers with < 1.0 alpha on the right)

However, what we can do instead is clear the canvas with a solid white color instead of transparent before drawing the layer's contents as normal. When blending the resulting bitmap over an opaque background using `PorterDuff.MULTIPLY` (i.e. `BlendModeCompat.MODULATE`), the end result will be as if we had used `BlendModeCompat.MULTIPLY`, since all-1.0 (white) is a multiplication identity:

Correct-Blend

This PR implements the latter solution and adds a consistent support for the Multiply blend mode for all Android versions.

*Test file used:*  [blendmode-tests-multiply+screen+bg.zip](https://github.com/user-attachments/files/16365843/blendmode-tests-multiply%2Bscreen%2Bbg.zip)

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
@@ -315,7 +315,22 @@ public abstract class BaseLayer
       }
 
       // Clear the off screen buffer. This is necessary for some phones.
-      clearCanvas(canvas);
+      if (getBlendMode() != LBlendMode.MULTIPLY) {
+        clearCanvas(canvas);
+      } else {
+        // Due to the difference between PorterDuffMode.MULTIPLY (which we use for compatibility
+        // with Android < Q) and BlendMode.MULTIPLY (which is the correct, alpha-blended mode),
+        // we will alpha-blend the contents of this layer on top of a white background before
+        // we multiply it with the opaque substrate below (with canvas.restore()).
+        //
+        // Since white is the identity color for multiplication, this will behave as if we
+        // had correctly performed an alpha-blended multiply (such as BlendMode.MULTIPLY), but
+        // will work pre-Q as well.
+        Paint solidWhite = new Paint();
+        solidWhite.setColor(0xffffffff);
+        canvas.drawRect(rect.left - 1, rect.top - 1, rect.right + 1, rect.bottom + 1, solidWhite);
+      }
+
       if (L.isTraceEnabled()) {
         L.beginSection("Layer#drawLayer");
       }
--- a/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
@@ -118,6 +118,8 @@ public abstract class BaseLayer
   float blurMaskFilterRadius = 0f;
   @Nullable BlurMaskFilter blurMaskFilter;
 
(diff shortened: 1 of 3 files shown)


gpeal (approved):
Thanks!

lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
```diff
@@ -315,7 +315,22 @@ public void draw(Canvas canvas, Matrix parentMatrix, int parentAlpha) {
...
+      } else {
+        // Due to the difference between PorterDuffMode.MULTIPLY (which we use for compatibility
+        // with Android < Q) and BlendMode.MULTIPLY (which is the correct, alpha-blended mode),
+        // we will alpha-blend the contents of this layer on top of a white background before
+        // we multiply it with the opaque substrate below (with canvas.restore()).
+        //
+        // Since white is the identity color for multiplication, this will behave as if we
+        // had correctly performed an alpha-blended multiply (such as BlendMode.MULTIPLY), but
+        // will work pre-Q as well.
+        Paint solidWhite = new Paint();
```
gpeal, line 329: Could you create this as a nullable class field that you set if null? That way it gets created once and lazily? You can also use `LPaint()` instead which is mostly the same but does slightly fewer allocations for things that aren't needed.

geomaster: Sure thing!

geomaster:
@gpeal Thanks for the review! Addressed comment. Let me know if there's anything else you see that needs improvement!

gpeal (approved):
Thanks for the investigative work here!

gpeal (approved):

PR #2523, opened 2024-07-24
Improve drop shadow effect accuracy

allenchen1154:
These changes make several improvements to how drop shadow effects are displayed:
- Drop shadows now take parent alpha into account. This means that if a layer or fill has an animated opacity, the drop shadow will multiply that opacity with its own.
- Adds drop shadow support to Image layers. There are some visual bugs with how the shadows are rendered, however.
- Applies a constant scale factor to the distance and softness values from the Lottie file before passing them to `Paint.setShadowLayer()` to more closely match how they are displayed in After Effects. See https://github.com/airbnb/lottie-ios/pull/2175 for similar changes on iOS.
- Adds three snapshot test files for distance, softness, and alpha validation.

Distance test file:

Softness test file:

Example of Image layer shadow bug, right side is how it should look (capture from After Effects):
image


lottie/src/main/java/com/airbnb/lottie/parser/DropShadowEffectParser.java
```diff
@@ -67,10 +67,10 @@ private void maybeParseInnerEffect(JsonReader reader, LottieComposition composit
               direction = AnimatableValueParser.parseFloat(reader, composition, false);
               break;
             case "Distance":
-              distance = AnimatableValueParser.parseFloat(reader, composition);
+              distance = AnimatableValueParser.parseFloat(reader, composition, false);
               break;
             case "Softness":
-              radius = AnimatableValueParser.parseFloat(reader, composition);
+              radius = AnimatableValueParser.parseFloat(reader, composition, false);
```
allenchen1154, line 73: I removed the dp -> px scale factor here since it's not clear to me that the units used by After Effects map to one or the other.

gpeal: Just did some tests and I think you're right here

lottie/src/main/java/com/airbnb/lottie/animation/keyframe/DropShadowKeyframeAnimation.java
```diff
@@ -2,6 +2,7 @@
 
 import android.graphics.Color;
 import android.graphics.Paint;
+import android.util.Log;
```
gpeal, line 5: Can you remove this import?

lottie/src/main/java/com/airbnb/lottie/animation/keyframe/DropShadowKeyframeAnimation.java
```diff
@@ -52,14 +53,21 @@ public void applyTo(Paint paint) {
     }
     isDirty = false;
 
+    // We scale the distance and softness value from the file so that the Paint.setShadowLayer() call
+    // gives results that more closely match After Effects
+    float AFTER_EFFECTS_DISTANCE_SCALE_FACTOR = 1.5f;
```
gpeal, line 58: Want to extract this to a const?

gpeal (commented):
Thanks for doing this investigation. How confident are you in these values? It might help to make an animation that has a drop shadow on a circle and a larger concentric stroked circle that represents where the shadow should end to ensure that it is correct. Have you tried anything like that?

allenchen1154:
@gpeal Good idea, for the distance scale factor we could use a animation that has a shadow with no blur, and use a reference point as you suggested that the shadow should touch at a predetermined distance. For the softness scaling we may need something like a paint swatch in the background?

gpeal:
> @gpeal Good idea, for the distance scale factor we could use a animation that has a shadow with no blur, and use a reference point as you suggested that the shadow should touch at a predetermined distance. For the softness scaling we may need something like a paint swatch in the background?

@allenchen1154 Yeah, something like that. Are you familiar enough with After Effects to make something like that?

allenchen1154:
I've added the test files we discussed and I'm reasonably confident in the scaling factors.

gpeal (approved):
This is great! Thanks for doing this. One small question around performance so let me know if you want to take a quick look at that but otherwise, LGTM!

gpeal:
Actually, after looking through the snapshot tests, are a few others that look a bit off/clipped as well. Do you want to take a look? I can send you any of the prod animations from the snapshot tests if you want any of them.

web
image
android
image


web
image
android
image

web (lottiefiles/the_final_frontier)
image
android
image

lottie/src/main/java/com/airbnb/lottie/animation/keyframe/DropShadowKeyframeAnimation.java
```diff
@@ -42,24 +46,27 @@ public DropShadowKeyframeAnimation(BaseKeyframeAnimation.AnimationListener liste
   }
 
   @Override public void onValueChanged() {
-    isDirty = true;
     listener.onValueChanged();
   }
 
-  public void applyTo(Paint paint) {
-    if (!isDirty) {
```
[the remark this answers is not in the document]
allenchen1154: @gpeal Would that look something like this?
```java
    // If the inputs haven't changed, we don't need to update the Paint
    if (direction.getValue() == lastAppliedDirection &&
        distance.getValue() == lastAppliedDistance &&
        radius.getValue() == lastAppliedRadius &&
        color.getValue() == lastAppliedColor &&
        opacity.getValue() == lastAppliedOpacity &&
        parentAlpha == lastAppliedParentAlpha) {
      return;
    }
```

(Assuming the `BaseKeyframeAnimation.getValue()` calls are cached and not expensive)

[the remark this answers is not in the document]
gpeal: Yup! The keyframe getValue() calls are designed to be pretty efficient but you can create local variables for them so that if you make it past the if, you can reuse the locals.

PR #2527, opened 2024-08-04
LottieCompositionFactory: Add factory methods that take an okio Source.

vanniktech:
Screenshot 2024-08-04 at 11 33 08

I can't use JsonReader, but since I'm anyway coming from a source, I don't mind and would like to use it right away.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -48,6 +48,7 @@ import java.util.zip.ZipInputStream;
 
 import okio.BufferedSource;
 import okio.Okio;
+import okio.Source;
 
 /**
  * Helpers to create or cache a LottieComposition.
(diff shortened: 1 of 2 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -402,26 +403,41 @@ public static LottieTask<LottieComposition> fromJsonString(final String json, @N
   @WorkerThread
   public static LottieResult<LottieComposition> fromJsonStringSync(String json, @Nullable String cacheKey) {
     ByteArrayInputStream stream = new ByteArrayInputStream(json.getBytes());
-    return fromJsonReaderSync(JsonReader.of(buffer(source(stream))), cacheKey);
+    return fromSourceSync(source(stream), cacheKey);
+  }
+
+  public static LottieTask<LottieComposition> fromSource(final Source source, @Nullable final String cacheKey) {
```
gpeal, line 409: What do you think about naming this `fromJsonSource` to make it clear that this should be a json file, not zip, gzip, etc?

vanniktech: I like it. I've tried to see whether I could do the magic bytes check just like you do in other methods but with that Context parameter it gets awkward which is needed for fonts in case they are embedded.

PR #2530, opened 2024-08-05
LottieCompositionFactory: Avoid NPE when animation contains a Font Family and Context is null.

vanniktech:


lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
```diff
@@ -538,7 +538,7 @@ public static LottieResult<LottieComposition> fromZipStreamSync(@Nullable Contex
   }
 
   @WorkerThread
-  private static LottieResult<LottieComposition> fromZipStreamSyncInternal(Context context, ZipInputStream inputStream, @Nullable String cacheKey) {
+  private static LottieResult<LottieComposition> fromZipStreamSyncInternal(@Nullable Context context, ZipInputStream inputStream, @Nullable String cacheKey) {
```
gpeal, line 541: My concern here is that a lot of animations are loaded from the server and this may lead to runtime crashes where the engineer doesn't pass context and then somebody else at some point in the future sends an animation that has fonts and then it leads to crashes. It's a tough situation to recover from because they can fix it but won't be able to add the animation because it'll continue to crash older apps.

Given that this can be an application context, what is the use case in which you can't get access to at least the application context?

vanniktech: This crash would also happen right now with the current code that we have since the line after:

```kotlin
File tempFile = new File(context.getCacheDir(), fileName);
```

would just crash with a NPE. I just saw this and wanted to improve the behavior by at least telling you why it didn't succeed. Do you want me to `throw` instead of returning a `LottieResult`?

gpeal: I see. IIRC, the intention here was that context is `@NonNull`, There is an overloaded version that omits the context parameter and clearly states in the docs that fonts will not be loaded.

vanniktech: > clearly states in the docs that fonts will not be loaded.

This isn't the case though as it will crash.

gpeal (approved):

PR #2531, opened 2024-08-05
Apply transformation for the stroke on the canvas so the skew is also…

Mirko-Volkers:
I noticed that the Android implementation had trouble with applying a skew on a stroke, so this attempts to fix that issue.

The issue was that since the transformation was applied to the path, it didn’t take any properties from the paint into consideration. As a result, the skew was not applied to the width of the line. By applying the transformation on the canvas, this issue is resolved.

Example Lottie which wasn’t rendered correctly (should have skewed lines):
[Skewed-Stroke.json](https://github.com/user-attachments/files/16495718/Skewed-Stroke.json)


gpeal:
@Mirko-Volkers Thanks for taking a look at this. Could you add `Skewed-Stroke.json` to [this](https://github.com/airbnb/lottie-android/tree/master/snapshot-tests/src/main/assets/Tests) folder? It'll get auto-included in snapshot tests if you do.

gpeal:
@Mirko-Volkers Based on the snapshot tests, it doesn't look like these changes are correct. Several of the failing ones are in the same snapshot-test module so feel free to take a look at them locally to see what's going on.

Mirko-Volkers:
Yeah, I see some differences in the snapshots. 
Found some changes because the scale was applied twice, going through it locally to check if I can fix it and make sure I'm not missing something.

gpeal (approved):
I just looked through the diffs and they look good to me. They're all either small aliasing differences or actually seem more accurate. Thanks for looking into this!

PR #2536, opened 2024-08-13
[a11y] Changing logic of how reduce motion options are set to match it with lottie iOS

pranayairan:
This PR introduces a new way to determine if reduce motion is enabled or not. It matches iOS Lottie API https://github.com/airbnb/lottie-ios/pull/2110. 

This change provides the ability for apps integrating Lottie to have custom reduce motion behavior which doesn't need to rely solely on system settings.

This change also allows reduce motion to work regardless of usage of LottieAnimationView, LottieAnimation in compose or straight LottieDrawable. 

Testing

- Verified that the default ReducedMotionOptionProvider is correctly used when none is provided.
- Ensured that custom implementations of ReducedMotionOption can be set and used.  
- Updated existing test for LottieDrawable reduce motion.


lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
```diff
@@ -371,17 +362,6 @@ private void init(@Nullable AttributeSet attrs, @AttrRes int defStyleAttr) {
...
   }
 
-  /**
-   * Allows ignoring system animations settings, therefore allowing animations to run even if they are disabled.
-   * <p>
-   * Defaults to false.
-   *
-   * @param ignore if true animations will run even when they are disabled in the system settings.
-   */
-  public void setIgnoreDisabledSystemAnimations(boolean ignore) {
```
gpeal, line 381: Removing an existing public API is tricky. At the very least, we'll need to deprecate this and provide a graceful fallback. Ideally, its behavior would remain the same (scope IgnoreDisabledSystemAnimations() to this view if possible).

It can be fully removed after a few releases.

gpeal (commented):
Thanks for looking into this!

I'm not totally sure all of the different ways this is used today. It's possible that somebody is intentionally using this on a per-view basis and this PR removes the ability to control it at that level.

Paired with my other comment about backwards compatibility, it would be good to potentially solve both with a setting at the view level that falls back to the global config when not set.

lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
pranayairan: @gpeal make sense, added the original API back.

pranayairan:
@gpeal wondering if this looks good can we merge it ?

lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -1244,7 +1245,8 @@ boolean isAnimatingOrWillAnimateOnVisible() {
   }
 
   private boolean animationsEnabled() {
-    return systemAnimationsEnabled || ignoreSystemAnimationsDisabled;
+    return (systemAnimationsEnabled || ignoreSystemAnimationsDisabled) &&
+        L.getReducedMotionOption().getCurrentReducedMotionMode(getContext()) == ReducedMotionMode.STANDARD_MOTION;
```
gpeal, line 1249: If I'm reading this correctly, if you set `ignoreSystemAnimationsDisabled`  but the system scale is 0 then:
* You would expect animationsEnabled to return true
* systemAnimationsEnable: false
* ignoreSystemAnimationsDisabled: true
* LottieConfig: REDUCED

The predicate becomes (false || true) && (REDUCED == STANDARD) which would be false.

I think this may also read easier if you treat `ignoreSystemAnimationsDisabled == true` as higher priority with:
```
if (ignoreSystemAnimationsDisabled) {
    return true;
}
return L.getReducedMotionOption().getCurrentReducedMotionMode(getContext()) == ReducedMotionMode.STANDARD_MOTION;
```

Just off the top of my head, I think this is correct and easier to read. Is that right?

pranayairan: you are right. Ideally `getCurrentReducedMotionMode` should not even be evaluated when `ignoreSystemAnimationsDisabled` is set to true. Let me fix it.

pranayairan: I changed it to this 

```
private boolean animationsEnabled() {
    if (ignoreSystemAnimationsDisabled) {
      return true;
    }
    return systemAnimationsEnabled &&
        L.getReducedMotionOption().getCurrentReducedMotionMode(getContext()) == ReducedMotionMode.STANDARD_MOTION;
  }
  ```
  We still need to consider old `systemAnimationsEnabled` flag that can be set on drawable directly for backward compatibility

**Review by gpeal — APPROVED** [review]
Thanks for working through this!

**Comment by pranayairan** [comment]
@gpeal thanks for merging. Can you share what is the release cadence we follow ? this will help me plan downstream changes.

**Comment by gpeal** [comment]
> @gpeal thanks for merging. Can you share what is the release cadence we follow ? this will help me plan downstream changes.

I just release 6.5.1 with your change!

**Comment by pranayairan** [comment]
@gpeal thanks. I am testing the integration and I think i miss something. The logic is not working when i am using `LottieAnimation` in compose. I am not entirely sure how this was working before since `LottieAnimation` doesn't call playAnimation on LottieDrawable, so none of the logic we have added previously and modified in this PR executes. 

From what i can understand looking into the code, when using `LottieAnimation` we directly draw on canvas.

**Comment by gpeal** [comment]
> @gpeal thanks. I am testing the integration and I think i miss something. The logic is not working when i am using `LottieAnimation` in compose. I am not entirely sure how this was working before since `LottieAnimation` doesn't call playAnimation on LottieDrawable, so none of the logic we have added previously and modified in this PR executes.
> 
> From what i can understand looking into the code, when using `LottieAnimation` we directly draw on canvas.

That's correct. I should have remembered to check for that in your PR. lottie-compose uses LottieDrawable but uses its own animator.

**Comment by pranayairan** [comment]
make sense, let me try creating a new PR for compose then.

**Comment by pranayairan** [comment]
@gpeal created another PR just for compose. https://github.com/airbnb/lottie-android/pull/2542

**Comment by tevjef** [comment]
Heads up our UI tests caught this crash this morning in `6.5.1`. I will create a issue once I've collected more info. 

```
 java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.ContentResolver android.content.Context.getContentResolver()' on a null object reference
 	at com.airbnb.lottie.utils.Utils.getAnimationScale(Utils.java:267)
 	at com.airbnb.lottie.configurations.reducemotion.SystemReducedMotionOption.getCurrentReducedMotionMode(SystemReducedMotionOption.java:22)
 	at com.airbnb.lottie.LottieDrawable.animationsEnabled(LottieDrawable.java:1252)
 	at com.airbnb.lottie.LottieDrawable.playAnimation(LottieDrawable.java:826)
```

PR #2540, opened 2024-08-27
Fix value callback is not called for PathKeyframeAnimation

alex-dorokhov:
A value callback was not called, when only single value is defined for keyframe in PathKeyframeAnimation.

This PR also adds the check for startValue and endValue at start, same way it is done for other types of KeyframeAnimation.


lottie/src/main/java/com/airbnb/lottie/animation/keyframe/PathKeyframeAnimation.java
```diff
@@ -20,11 +20,12 @@ public PathKeyframeAnimation(List<? extends Keyframe<PointF>> keyframes) {
   }
 
   @Override public PointF getValue(Keyframe<PointF> keyframe, float keyframeProgress) {
+    if (keyframe.startValue == null || keyframe.endValue == null) {
```
gpeal, line 23: Is this strictly necessary? I'm afraid that this might cause backwards compatibility issues. Companies sometimes try to minify their animations in unexpected ways and I could imagine this crashing for animations that used to work.

gpeal (approved):

PR #2541, opened 2024-08-28
Apply scaling factor to drop shadow softness

allenchen1154:
Directly using the "Softness" value from the Lottie file (after accounting for dp and layer scaling) as the `radius` argument to `Paint.setShadowLayer()` results in shadows that are too soft. Since the underlying implementation calls into JNI code and is undocumented, we can only take a best guess as to how the After Effects softness value should be mapped to this argument. The [Lottie web implementation](https://github.com/airbnb/lottie-web/blob/master/player/js/elements/svgElements/effects/SVGDropShadowEffect.js#L63) multiplies the value by a constant factor of 0.25, so we take a similar approach here, using a value of 0.33 to achieve a close match to a reference image.


gpeal (approved):
Thanks for taking another look here!

PR #2544, opened 2024-08-30
Fix resolving key path for matte layers

alex-dorokhov:
This PR adds the check for matte layer name before going deeper during key path resolution.

Without the fix the key path matches children inside the matte layer even though it should not, for example:

'layer_1.group_1.child_1' could match 'layer_1.**matte_1**.group_1.child_1'

With the fix it is not possible anymore.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/model/layer/BaseLayer.java
@@ -746,7 +746,7 @@ public abstract class BaseLayer
         accumulator.add(matteCurrentPartialKeyPath.resolve(matteLayer));
       }
 
-      if (keyPath.propagateToChildren(getName(), depth)) {
+      if (keyPath.matches(matteLayer.getName(), depth) && keyPath.propagateToChildren(getName(), depth)) {
         int newDepth = depth + keyPath.incrementDepthBy(matteLayer.getName(), depth);
         matteLayer.resolveChildKeyPath(keyPath, newDepth, accumulator, matteCurrentPartialKeyPath);
       }


gpeal (commented):
Could you add a snapshot test for this? You can follow other test cases [here](https://github.com/airbnb/lottie-android/blob/master/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/tests/DynamicPropertiesTestCase.kt) as an example.

It may be difficult to fully run the tests locally because it requires API keys but you can comment out other test cases and then put a breakpoint [here](https://github.com/airbnb/lottie-android/blob/a1fb304f928be465ffcb344c3952cd47325ba48d/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/SnapshotTestCaseContext.kt#L123) to view the bitmap snapshot.

gpeal:
@alex-dorokhov would you like to update this so we can get this PR landed?

alex-dorokhov:
> @alex-dorokhov would you like to update this so we can get this PR landed?

Sorry, that it got so long for me to reply.

There is already a [test](https://github.com/airbnb/lottie-android/blob/master/snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/tests/DynamicPropertiesTestCase.kt#L471) for dynamic properties of matte layers. I believe it had wrong results before the fix and now it is good, so should be accepted. Please, look at the second report here https://github.com/airbnb/lottie-android/pull/2544#issuecomment-2322164029

I've debugged the test to be sure.

### Before the fix

This line has called twice when callback is added (so two matte layers have been matched).

Screenshot 2024-10-04 at 19 13 23

Please, notice that "Shape Layer 3" should not be matched:

Screenshot 2024-10-04 at 19 13 08

### After the fix

This line has been called only with "Shape Layer 1":

Screenshot 2024-10-04 at 19 14 09

I've also attached **before** and **after** videos for convenience: 

https://github.com/user-attachments/assets/3cdc3d2a-84d3-4000-b57f-cd9060b7580d

https://github.com/user-attachments/assets/ef0332a6-64c9-4773-ad4a-dd48c2992c36

gpeal (approved):

gpeal:
@alex-dorokhov Thanks for clarifying!

PR #2548, opened 2024-09-13
Drop shadow overhaul: improve correctness and performance

geomaster:
## High-level summary

This PR introduces a large change to how drop shadows are rendered, introducing an `applyShadowsToLayers` flag which, by analogy to `applyOpacitiesToLayers`, allows layers to be treated as a whole for the purposes of drop shadows, improving the accuracy and bringing lottie-android in line with other renderers (lottie-web and lottie-ios).

Several different codepaths for different hardware/software combinations are introduced to ensure the fastest rendering available, even on legacy devices.

The calculation of shadow direction with respect to transforms is improved so that the output matches lottie-web and lottie-ios.

Image layers now cast shadows correctly thanks to a workaround to device-specific issues when combining `Paint.setShadowLayer()` and bitmap rendering.

Even in non-`applyShadowsToLayers` mode, correctness is improved by allowing the shadow-to-be-applied to propagate in a similar way as alpha. This allows some amount of visual fidelity to be recovered for animations or environments where enabling `applyShadowsToLayers` is not possible.

A number of issues that caused incorrect rendering in some other cases have been fixed.

### Drop shadows in Lottie

Lottie specifies drop shadows as a tuple of (angle, distance, radius, color, alpha), with each element being animatable.

The consensus behavior for the rendering of a layer with a drop shadow, which seems to be mostly respected in lottie-web and lottie-ios, seems to be:

1. Evaluate the values at the current frame for angle (`theta`), distance (`d`), radius (`r`), color with alpha (`C`).
2. Apply the layer transform and render the layer normally to a surface `So` (original layer).
3. Copy `So` to new surface `Ss` (shadow).
4. Apply a gaussian blur of radius `r' = c * r` to `Ss`, where `c` is some platform-specific constant intended to normalize blur implementations between platforms. (Ours is 0.33, lottie-web's is 0.25; see https://github.com/airbnb/lottie-android/pull/2541).
5. Tint `Ss` with the color and combine the alpha by applying the following for each pixel `P`: `P.rgb = C.rgb * P.a; P.a = C.a * P.a`.
6. Now the shadow is ready on `Ss`, and needs to be drawn into its final position.
7. Convert from polar coordinates `theta` and `d` into `dx` and `dy`, with the 0 position at 12 o'clock: `dx = d * cos(theta - pi/2); dy = d*sin(theta - pi/2)`.
8. Draw `Ss` onto `Si` (intermediate surface) with a translation of `(dx, dy)`.
9. Draw `So` (original layer) onto `Si` with identity transform.
10. Compose `Si` into the framebuffer using the layer's alpha and blend mode.

Some non-obvious consequences of the definition above:
- The angle, distance, and radius are relative to the layer post-transform, not pre-transform. That is, rotating the layer (via its transform) still keeps the same screen-space direction of the shadow, and scaling the layer (via its transform) still keeps the same screen-space shadow blur radius.
- The drop shadow is not based on any derived outline, so a layer's drop shadow can be seen through its non-fully-opaque pixels. At the same time, reducing the alpha of a pixel in a layer reduces its alpha in the drop shadow.
- A layer's shadow and the layer do not blend on top of each other on the final canvas in case the layer has a blend mode or alpha. Instead, the shadow and the layer are alpha-blended with each other, and the result is then composited onto the canvas.
  - In case the layer has a normal blend mode, this is equivalent to alpha-blending the layer's shadow and then the shadow onto the canvas separately.

### Drop shadows in lottie-android currently

lottie-android's current implementation of drop shadows differs in important ways:
1. **Shadows are applied per-shape.** This means that a case like a shape with both fill and stroke has incorrect shadows, since both the fill and the stroke render a separate shadow on top of each other.
2. **Precomp layer shadows are ignored.** This means that a precomp cannot cause any of its child shapes to cast a shadow. This is a consequence of the current implementation of (1).
3. **Image layers do not render correct shadows,** due to the minefield that is the support matrix (or in Android's case, a more apt name would be a support tensor) of Android's graphics stack - `setShadowLayer()` simply doesn't work for images consistently. (See the last image in https://github.com/airbnb/lottie-android/pull/2523#issue-2428578510.)

## Contributions of this PR

This PR introduces the following improvements and additions.

1. **Move the drop shadow model from individual content elements to layers,** and add some missing keypath callbacks. This is a prerequisite for handling drop shadows on a layer level.
2. **An `OffscreenLayer` implementation,** which serves as an abstraction that can replace `canvas.saveLayer()` for off-screen rendering and composition onto the final bitmap, but with the important distinction that it can also handle drop shadows, and possibly use hardware-accelerated `RenderNode`s and `RenderEffects` where available.
    - To use an `OffscreenLayer`, call its `.start()` method with a parent canvas and a `ComposeOp`, and draw on the *returned canvas.* Once finished, call `OffscreenLayer.finish()` to compose everything from the returned canvas to the parent canvas, applying alpha, blend mode, drop shadows, and color filters.
    - `OffscreenLayer` makes a dynamic decision on what to use for rendering - a no-op, forward to `.saveLayer()`, a HW-accelerated `RenderNode`, or a software bitmap, depending on the requested `ComposeOp` and hardware/SDK support.
    - The hope is that `OffscreenLayer` becomes a useful abstraction that can be extended to e.g. support hardware blurs, multiple drop shadows, or to support mattes in a hardware-accelerated fashion where possible. 
3. **The `applyShadowsToLayers` flag** which, by analogy to `applyOpacityToLayers`, turns on a more accurate mode that implements the drop shadow algorithm described above.
    - `OffscreenLayer` is used to apply alpha if `applyOpacityToLayers` is enabled, and to apply shadows if `applyShadowsToLayers` is enabled. The cost is paid only once if both alpha and drop shadows are present on a layer.
    - Not all `saveLayer()` calls in the code have been rewritten to use `OffscreenLayer` - the blast radius is minimized. `OffscreenLayer` is presently used only to apply alpha and drop shadows, and blend mode and color filters are still applied in `BaseLayer` using `saveLayer()` directly.
4. **More accurate shadow transformations.** Previously, the angle and distance were pre-transform, and only the radius was post-transform (contrary to step (2) of the algorithm). We correct this to match other renderers.
5. **More complete shadow handling even when `applyShadowsToLayers` is `false`:** we plumb the shadow through `.draw()` and `drawLayer()` calls similarly to alpha, and this allows us to render per-shape shadows on children of composition layers too.
6. ***Workaround for drop shadows on image layers.**
    - The workaround relies on `OffscreenLayer` as well, and image layers now render shadows properly in all cases.
7. **Fixes to a few subtle issues** causing incorrect rendering in other cases. (will be marked using PR comments, I might have forgotten some)

## Open questions

* **Should `applyShadowsToLayers` be `true` by default?** Some codepaths, such as when rendering purely via software, can be slow if shadow-casting layers are exceedingly large. But, the performance is still acceptable, and in the vast majority of cases everything is quite snappy.
* **Have I introduced any regressions?** The snapshot tests should answer this.
* **How does this perform on older devices?** `applyShadowsToLayers` plus an old device should trigger the purely-software shadow rendering mode. Simulating this in condition manually yields accurate results, and the performance seems surprisingly good, but it's unclear what will happen on a lower-end phone. There's also always the possibility of some device subtlety being missed. I don't have access to an older Android device.

## Testcases

These files now match between lottie-web and lottie-android:

[drop_shadow_comparator.json](https://github.com/user-attachments/files/16997070/drop_shadow_comparator.json)

[simple_shadow_casters_ll2.json](https://github.com/user-attachments/files/16997084/simple_shadow_casters_ll2.json)

The files from this earlier PR still all render the same: https://github.com/airbnb/lottie-android/pull/2523, with the exception of the fix for image layer bug, which fixes the rendering of the Map icon as mentioned in the comment of that PR.

This file has been used as a perf stress test with many <255 opacity precomps, some stacked inside each other, that must all be blended separately: [precomp_opacity_killer.json](https://github.com/user-attachments/files/16997261/precomp_opacity_killer.json)


lottie/src/main/java/com/airbnb/lottie/animation/content/BaseStrokeContent.java
```diff
@@ -164,8 +160,10 @@ public abstract class BaseStrokeContent
       }
       return;
     }
-    int alpha = (int) ((parentAlpha / 255f * ((IntegerKeyframeAnimation) opacityAnimation).getIntValue() / 100f) * 255);
-    paint.setAlpha(clamp(alpha, 0, 255));
+    float strokeAlpha = opacityAnimation.getValue() / 100f;
+    int alpha = (int) (parentAlpha * strokeAlpha);
+    alpha = clamp(alpha, 0, 255);
```
geomaster, line 165: Applying this clamp right away, here and elsewhere, fixes issues with >255 or <0 alpha values, which caused shadows to not appear.

Such alpha values are possible when interpolating using an easing that evaluates to <0 or >1 at some point along its curve.

lottie/src/main/java/com/airbnb/lottie/animation/content/ContentGroup.java
```diff
@@ -175,24 +178,40 @@ Matrix getTransformationMatrix() {
...
+        (shadowToApply != null && lottieDrawable.isApplyingShadowToLayersEnabled() && hasTwoOrMoreDrawableContent());
+    int childAlpha = isRenderingWithOffScreen ? 255 : layerAlpha;
+
+    Canvas contentCanvas = canvas;
     if (isRenderingWithOffScreen) {
       offScreenRectF.set(0, 0, 0, 0);
-      getBounds(offScreenRectF, matrix, true);
-      offScreenPaint.setAlpha(layerAlpha);
-      Utils.saveLayerCompat(canvas, offScreenRectF, offScreenPaint);
+      getBounds(offScreenRectF, parentMatrix, true);
```
geomaster, line 189: This one got lost in the diff, but is very important - `matrix` was being passed here, causing incorrect transformations being applied when `applyOpacitiesToLayers` is true; see the `precomp_opacity_killer.json` testcase.

lottie/src/main/res/values/attrs.xml
```diff
@@ -17,6 +17,8 @@
         <attr name="lottie_imageAssetsFolder" format="string" />
         <attr name="lottie_progress" format="float" />
         <attr name="lottie_enableMergePathsForKitKatAndAbove" format="boolean" />
+        <attr name="lottie_applyOpacityToLayers" format="boolean" />
```
geomaster, line 20: this was missing for `applyOpacityToLayers`

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+  public Canvas start(Canvas parentCanvas, RectF bounds, ComposeOp op) {
+    if (this.parentCanvas != null) {
+      throw new RuntimeException("Cannot nest start() calls on a single OffscreenBitmap - call finish() first");
+    }
+
+    // Determine the scaling applied by the parentCanvas' pre-existing transform matrix. This is an optimization
+    // to avoid creating bitmaps (or render nodes) with unreasonable sizes that will get scaled down when drawn
+    // onto parentCanvas anyhow.
+    if (preExistingTransform == null) preExistingTransform = new float[9];
+    parentCanvas.getMatrix().getValues(preExistingTransform);
```
gpeal, line 217: getMatrix() allocates a new Matrix (and float array) and copies the values. To avoid allocations on the hot path, can you create a final field Matrix that you use to copy values to instead?

(same for the other getMatrix() calls)

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+ *
+ * In this sense, using an OffscreenLayer is very similar to the Canvas.saveLayer() family
+ * of functions, and in fact forwards to Canvas.saveLayer() when appropriate.
+ *
+ * Unlike Canvas.saveLayer(), an OffscreenLayer also supports compositing with a drop-shadow,
+ * and uses a hardware-accelerated target when available. It attempts to choose the fastest
+ * approach to render and composite the contents, up to and including simply rendering
+ * directly, if possible.
+ */
+public class OffscreenLayer {
```
gpeal, line 40: When I checked out this PR, Android Studio warns of a bunch of nullability issues. They're probably fine in practice but I aim to have all files have no warnings in Android Studio so that new issues are easy to spot. Could you try to tweak the code so that there aren't the nullability issues?

Often, it happens when you do something like:
```
@Nullable Bitmap bitmp…

void foo() {
    if (bitmap != null) {
        bitmap.something()
    }
```

Another thread could technically set bitmap to null so I would do:

```
void foo() {
    Bitmap bitmap = this.bitmap;
    if (bitmap != null) …
}
```

That way your local copy is guaranteed to not change.

lottie/src/main/java/com/airbnb/lottie/utils/DropShadow.java
```diff
@@ -0,0 +1,128 @@
...
+  public void setColor(int color) {
+    this.color = color;
+  }
+
+  public boolean sameAs(DropShadow other) {
+    return radius == other.radius && dx == other.dx && dy == other.dy && color == other.color;
+  }
+
+  public void transformBy(Matrix matrix) {
+    float[] vecs = {dx, dy};
```
gpeal, line 71: Could you create a field so we don't allocate the float array in the hot path?

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+  @Nullable private RenderNode shadowRenderNode; // Render node for the shadow
+  @Nullable private Canvas renderNodeCanvas;
+  @Nullable private DropShadow lastRenderNodeShadow;
+
+  public OffscreenLayer() {
+    this.clearPaint = new LPaint();
+    this.clearPaint.setXfermode(new PorterDuffXfermode(PorterDuff.Mode.CLEAR));
+  }
+
+  private RenderStrategy chooseRenderStrategy(Canvas parentCanvas, RectF bounds, ComposeOp op) {
```
gpeal, line 145: bounds isn't used

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+  @Nullable private Canvas shadowBitmapCanvas;
+  @Nullable private Canvas shadowMaskBitmapCanvas;
+  @Nullable private LPaint shadowPaint;
+  @Nullable private BlurMaskFilter shadowBlurFilter;
+  private float lastShadowBlurRadius = 0.0f;
+
+  // For RenderStrategy.RENDER_NODE:
+  @Nullable private RenderNode renderNode; // Render node with the initial contents of the layer
+  @Nullable private RenderNode shadowRenderNode; // Render node for the shadow
+  @Nullable private Canvas renderNodeCanvas;
```
gpeal, line 137: This can be a local variable

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+
+    if (!op.hasShadow()) {
+      // Canvas.saveLayer() supports alpha-compositing, blend modes, and color filters, which is
+      // sufficient for this case, and is faster than manually maintaining an off-screen bitmap.
+      // It is not clear if it's faster than RENDER_NODE and when, but this is what we've been
+      // doing prior to OffscreenLayer, so keep that behavior to be safe and avoid regressions.
+      return RenderStrategy.SAVE_LAYER;
+    }
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q && parentCanvas.isHardwareAccelerated()) {
```
gpeal, line 159: nit: I like how you have the two guard cases above. I think having the BITMAP case at the end instead be another guard statement here, it would be easier to read. It'll also allow you un-nest this nested if.

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  private Bitmap allocateBitmap(RectF bounds, Bitmap.Config cfg) {
+    // Add 5% to the width and height before allocating. This avoids repeated
+    // reallocations in the worst cases of a bitmap growing every frame, while
+    // still being relatively speedy to blit and operate on.
+    int width = (int)Math.ceil(bounds.width() * 1.05);
+    int height = (int)Math.ceil(bounds.height() * 1.05);
+    Bitmap bmp = Bitmap.createBitmap(width, height, cfg);
+    return bmp;
```
gpeal, line 183:
```suggestion
    return Bitmap.createBitmap(width, height, cfg);
```

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+    if (bounds.width() < bitmap.getWidth() * 0.75f || bounds.height() < bitmap.getHeight() * 0.75f) {
+      return true;
+    }
+
+    return false;
+  }
+
+  public Canvas start(Canvas parentCanvas, RectF bounds, ComposeOp op) {
+    if (this.parentCanvas != null) {
+      throw new RuntimeException("Cannot nest start() calls on a single OffscreenBitmap - call finish() first");
```
gpeal, line 210:
```suggestion
      throw new IllegalStateException("Cannot nest start() calls on a single OffscreenBitmap - call finish() first");
```

RuntimeExceptions usually shouldn't be thrown directly. Their subclasses are more specific.

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,494 @@
...
+    // Determine the scaling applied by the parentCanvas' pre-existing transform matrix. This is an optimization
+    // to avoid creating bitmaps (or render nodes) with unreasonable sizes that will get scaled down when drawn
+    // onto parentCanvas anyhow.
+    if (preExistingTransform == null) preExistingTransform = new float[9];
+    parentCanvas.getMatrix().getValues(preExistingTransform);
+
+    float pixelScaleX = preExistingTransform[Matrix.MSCALE_X];
+    float pixelScaleY = preExistingTransform[Matrix.MSCALE_Y];
+
+    RectF scaledBounds = new RectF(
```
gpeal, line 222: Similar to above, could this be updated to not do any allocations in the hot path?

lottie/src/main/java/com/airbnb/lottie/animation/content/ContentGroup.java
```diff
@@ -175,24 +178,40 @@ Matrix getTransformationMatrix() {
...
+      offscreenOp.alpha = layerAlpha;
+      if (shadowToApply != null) {
+        shadowToApply.applyTo(offscreenOp);
+        shadowToApply = null; // Don't pass it to children - OffscreenLayer now takes care of this
```
gpeal, line 193: In all of my years writing Java, I always thought that parameters were final 😮

gpeal (commented):
Wow, this is really amazing work. I really appreciate it and also how this can directly lead to further improvements for mattes and masks.

I left a few comments but a couple of broad themes:
1. Take a look at the warnings in Android Studio and try to make them all go away. I try to keep the gutter "green" so that it's easy to spot new issues
2. There are a bunch of new allocations in the hot path. Calls to getMatrix() are one example but there are also RectFs, float arrays, etc. It would be great to keep the hot path clean of allocations
3. Run the Android Studio formatter on the new files to fix a few consistency issues

Looking forward to working with you to get this in!

gpeal:
Also, I'm not sure why the snapshot diff against master is so big but if you compare it to 453b43cbf0003ce182e37a72d9fc43b882d39187 which is at
https://happo.io/a/27/p/27/compare/453b43cbf0003ce182e37a72d9fc43b882d39187-android31/ca1b2040a05692731c919a90ad0c066615f43797-android31?t=diffs

You can see a much smaller snapshot diff. I just scanned it and the diffs look good to me! Feel free to scan them yourself as well.

9I73 (approved):

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
geomaster: I tried to fix all issues and I think now all of the warnings are pre-existing. Let me know if I missed something!

lottie/src/main/java/com/airbnb/lottie/animation/content/ContentGroup.java
geomaster: I'm to embarrassed to even start listing the surprising things I learned about programming languages I've used entirely too late :D

geomaster:
@gpeal Thanks for the review, Gabriel! Addressed all of the comments. The snapshot diffs look fine to me as well, but I'd also like to see another run with the latest round of changes, to ensure nothing was regressed accidentally.

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
geomaster: Done!

But I would sincerely hope that the runtime does some basic escape analysis and can decide to allocate objects like this on the stack. And even if not, I'd assume the GC pressure is a much smaller problem than the actual time we're taking to render the frame.

I wouldn't mention this normally, but there are places in the code where an allocation cannot be avoided (or it would be cumbersome), such as when you want to pass around a complex object. And the places where I did convert local variables to members to avoid allocations are much harder to read now, there are new nullability issues that need to be addressed, etc.

The optimist in me wants to say that the runtime *surely* has a way of dealing with this, given that Java doesn't really have anything like a `struct`, and this is basically the only option.

My memory is foggy, but even looking at major perf-sensitive Java codebases like Minecraft, I recall they don't shy away from per-frame allocations.

You probably have a much stronger intuition on this than me, so let me know your thoughts.

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -0,0 +1,544 @@
...
+        if (clearPaint == null) {
+          clearPaint = new LPaint();
+          clearPaint.setXfermode(new PorterDuffXfermode(PorterDuff.Mode.CLEAR));
+        }
+
+        if (needNewBitmap(bitmap, scaledBounds)) {
+          if (bitmap != null) {
+            deallocateBitmap(bitmap);
+          }
+          bitmap = allocateBitmap(scaledBounds, Bitmap.Config.ARGB_8888);
```
gpeal, line 274:
```suggestion
        if (needNewBitmap(bitmap, scaledBounds)) {
          if (bitmap != null) {
            deallocateBitmap(bitmap);
          }
          bitmap = allocateBitmap(scaledBounds, Bitmap.Config.ARGB_8888);
```

I think it would be great if you could encapsulate the `needNew` + `deallocate` + `allocate` into a single helper that returns the bitmap.

Then, in each of the functions that calls the new function (something like `getBitmap(Rect)`, it would just use the return value as a local variable rather than using fields on the object

gpeal (commented):
This is really great work! One small comment to clean up some code and then rebase and I'm excited to see the snapshot diffs.

Run `./gradlew apiDump` and it'll update the `.api` file and CI will pass. Again, I need to figure out a sustainable way to maintain abi backwards compatibility but I don't want to block this PR on that.

gpeal (approved):
This is truly fantastic work. I genuinely thank you for the hours you put into this. This is, without any doubt, the largest external contribution to this repo ever and it looks great. I hope we can build on this to get render node wins for mattes and masks.

PR #2559, opened 2024-09-29
Add an option to not clear the network cache

### LINKED ISSUE #2498: Option to exclude network cache from `LottieCompositionFactory.clearCache()`

In scenarios where Lottie animations are fetched from a URL and cached in the network cache for efficient loading, developers may encounter a consideration with the current implementation of the `clearCache()` method in `LottieCompositionFactory`. For example, during app startup, when a large Lottie animation is displayed on the splash screen, it gets cached in both the composition cache and the network cache. However, as this animation is typically only shown once and not reused during the app's lifecycle, there's an opportunity to optimize cache management.

A solution would be to extend `clearCache()` with an additional parameter that allows developers to specify whether they want to exclude the clearing of the network cache. By doing so, developers can selectively clear the composition cache while preserving the network cache for animations fetched from URLs.

I can’t think of another solution.

**Additional context**
Not applicable.

gpeal:
Fixes #2498

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -86,12 +86,27 @@ public class LottieCompositionFactory {
     LottieCompositionCache.getInstance().resize(size);
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


savvaspit:
Thank you @gpeal

PR #2571, opened 2024-11-04
Improve blur effect accuracy

geomaster:
Analogously to drop shadows in #2548, allow blurs to be applied to composition and image layers using the new `OffscreenLayer`.

Rename `applyShadowToLayers` to `applyEffectsToLayers` and use it for both of these effects. Keep externally-facing `applyShadowToLayers` and forward to `applyEffectsToLayers` for backwards compat.

Undo the layer transform when computing the radius of a blur effect, similarly as we do for shadows.

Use fast hardware blur using RenderEffect if possible, and otherwise delegate to a "reasonably fast" box-blur implementation on the CPU.

**This is not the final version. Still to do:**
* Try to optimize FastBlur more
* Check behavior on existing testcases to ensure no regressions
* Match the scaling constants between implementations to ensure the blur appears the same as in lottie-web, the same with `applyEffectsToLayers` and without, and the same with hardware vs software implementation of blur.

### DIFF
--- a/lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
+++ b/lottie-compose/src/main/java/com/airbnb/lottie/compose/LottieAnimation.kt
@@ -49,6 +49,11 @@ import kotlin.math.roundToInt
  *                             expense of performance.
  *                             Note: This process is very expensive. The performance impact will be reduced
  *                             when hardware acceleration is enabled.
+ * @param applyEffectsToLayers Sets whether to apply effects such as drop shadow and blur to each layer instead of
+ *                             individually per shape. In case where effects are applied to shape groups and
+ *                             composition layers, this will improve correctness but at a performance cost.
+ *                             Note: This process is expensive when hardware acceleration is unavailable.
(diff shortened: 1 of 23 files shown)


gpeal (approved):
This LGTM! We can merge through the API compat check because I don't have a good solution yet. Let me know if this is ready to merge on your end!

geomaster:
Thanks so much, @gpeal! Looking at some last remaining render diffs - I think I shouldn't have changed this rounding: https://github.com/airbnb/lottie-android/pull/2571/files#diff-f59159579cb60a5c96b48c67ce22f4efb9c861bac8d83d3a9361a63f5797a7a8R398, so will probably revert that back, check with another snapshot run and then I'm completely ok to merge. Thanks again for the quick eyes on this!

gpeal:
@geomaster Would you like to get this across the finish line?

geomaster:
Hey @gpeal, that would be good - I got caught up in a whirlwind of other work at the time and neglected this PR. I'll see if I can take a quick look in the following few days, I think it's pretty much ready. Thanks for reminding me!

Cherrera1980v (approved):

anandfresh (approved):

PR #2582, opened 2024-11-23
Refactoring a complex method, complex condition and a variable name

NeslynChellaiah:
Refactored a complex method onRestoreInstanceState and renamed a variable in the same method in LottieAnimationView.java. Refactored a complex conditional in ensureSoftwareRenderingBitmap method in LottieDrawable.java


gpeal:
Thanks for taking a look here but extracting these methods puts logic that should only be called from its original method into a method that could ambiguously be called at any time.

PR #2591, opened 2024-12-09
Binary compatibility with 6.5 and lower on LottieAnimation

### LINKED ISSUE #2568: Library update from 6.5.2 to 6.6.0 cause crash
After updating to 6.6.0 my app crashes with the error:
java.lang.NoSuchMethodError: No static method LottieAnimation(Lcom/airbnb/lottie/LottieComposition;Lkotlin/jvm/functions/Function0;Landroidx/compose/ui/Modifier;ZZZLcom/airbnb/lottie/RenderMode;ZLcom/airbnb/lottie/compose/LottieDynamicProperties;Landroidx/compose/ui/Alignment;Landroidx/compose/ui/layout/ContentScale;ZZLjava/util/Map;Lcom/airbnb/lottie/AsyncUpdates;ZLandroidx/compose/runtime/Composer;III)V in class Lcom/airbnb/lottie/compose/LottieAnimationKt; or its super classes (declaration of 'com.airbnb.lottie.compose.LottieAnimationKt'

works fine with 6.5.2

inorichi:
After #2548 was merged, the binary compatibility of the `LottieAnimation` composable changed, making it incompatible with previous versions. This PR fixes that.

Motivation: our team has a multi-repo approach and some colleagues did not update lottie for some reasons and we need to synchronize the update. With this PR we should not need to do that.

Fixes #2568


gpeal:
Unfortunately, this makes `LottieAnimation` appear as deprecated for everybody unless they use the new parameter.

This is an issue I'm aware but but don't know of any good solutions. This one won't work though, unfortuatenly.

Can you think of any others?

inorichi:
Can you check again? It shouldn't show up with the hidden deprecation level.

By the way, I only added compatibility to the first composable of the file, which is the method we're using. The second one is already deprecated so maybe it doesn't make sense to support it, but I'm not sure about the last one 🤔 .

gpeal:
> Can you check again? It shouldn't show up with the hidden deprecation level.
> 
> By the way, I only added compatibility to the first composable of the file, which is the method we're using. The second one is already deprecated so maybe it doesn't make sense to support it, but I'm not sure about the last one 🤔 .

Seems to be okay! Thanks for the update

gpeal (approved):

Issue #2271, opened 2023-03-31
App crash on invalid json url


vinodkumarsagitla:
My issue is when josn url exist on remote server it working fine but **not exist then app is crash** with below exception
But in **ios app not crash** (no animation is showing i.e empty view) with invalid same josn url  
I try to put code in **try-catch** but still app crash
So my question is their is any way to check is **url exist on server or not** or any way to prevent crash for url not exist

`java.lang.IllegalStateException: Unable to parse composition
                                                                              	at com.airbnb.lottie.LottieAnimationView.lambda$static$0(LottieAnimationView.java:72)
                                                                              	at com.airbnb.lottie.LottieAnimationView$$ExternalSyntheticLambda2.onResult(Unknown Source:2)
                                                                              	at com.airbnb.lottie.LottieAnimationView$1.onResult(LottieAnimationView.java:84)
                                                                              	at com.airbnb.lottie.LottieAnimationView$1.onResult(LottieAnimationView.java:77)
                                                                              	at com.airbnb.lottie.LottieTask.notifyFailureListeners(LottieTask.java:162)
                                                                              	at com.airbnb.lottie.LottieTask.lambda$notifyListeners$0$com-airbnb-lottie-LottieTask(LottieTask.java:138)
                                                                              	at com.airbnb.lottie.LottieTask$$ExternalSyntheticLambda0.run(Unknown Source:2)
                                                                              	at android.os.Handler.handleCallback(Handler.java:942)
                                                                              	at android.os.Handler.dispatchMessage(Handler.java:99)
                                                                              	at android.os.Looper.loopOnce(Looper.java:201)
                                                                              	at android.os.Looper.loop(Looper.java:288)
                                                                              	at android.app.ActivityThread.main(ActivityThread.java:7872)
                                                                              	at java.lang.reflect.Method.invoke(Native Method)
                                                                              	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
                                                                              	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
                                                                              Caused by: com.airbnb.lottie.parser.moshi.JsonEncodingException: Use JsonReader.setLenient(true) to accept malformed JSON at path $
                                                                              	at com.airbnb.lottie.parser.moshi.JsonReader.syntaxError(JsonReader.java:251)
[...]

PR #2616, opened 2025-03-02
remove unhandled exception in default failure listener

kunal26das:
Issue: https://github.com/airbnb/lottie-android/issues/2271

Description:
Currently if there is no failure listener being set to the `LottieAnimationView` the `DEFAULT_FAILURE_LISTENER` is used, which throws an unhandled exception on the main thread, which causes app to crash. There is no graceful handling in an event a developer forgets to set this explicitly. The same scenario isn't there for [lottie-ios](https://github.com/airbnb/lottie-ios). This is reproducible if by mistake an invalid url is passed to `LottieAnimationView` or a malformed lottie json asset is passed. Either way the component throws an unhandled exception and this is a big issue with production grade apps.

Solution:
As a fix, in the `DEFAULT_FAILURE_LISTENER` logging the exception instead of throwing it. Have also made the top level `failureListener` as non-null so that developers won't accidentally replace the `DEFAULT_FAILURE_LISTENER`.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -112,7 +112,7 @@ import java.util.zip.ZipInputStream;
       if (targetView.fallbackResource != 0) {
         targetView.setImageResource(targetView.fallbackResource);
       }
-      LottieListener<Throwable> l = targetView.failureListener == null ? DEFAULT_FAILURE_LISTENER : targetView.failureListener;
+      LottieListener<Throwable> l = targetView.failureListener;
       l.onResult(result);
     }
   }
--- a/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java
@@ -71,7 +71,7 @@ import java.util.zip.ZipInputStream;
       Logger.warning("Unable to load composition.", throwable);
       return;
     }
-    throw new IllegalStateException("Unable to parse composition", throwable);
+    Logger.warning("Unable to parse composition", throwable);
   };
 
   private final LottieListener<LottieComposition> loadedListener = new WeakSuccessListener(this);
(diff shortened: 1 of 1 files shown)


gpeal:
I originally chose to fail loudly by design here. My assumption is that most teams will notice if they don't gracefully handle errors and their app crashes but may never notice if their app is silently failing for some users indefinitely.

Ultimately, I slightly regret having any sort of networking built into the library at all because doing it correctly requires you to think about error handling to a degree that is probably out of scope for what the core Lottie library should be handling. I do recognize the convenience of it though.

Given that, I think failing loudly is actually better here because you take the short term hit if you don't properly handle it but it forces you to get things right in the long run which is better.

kunal26das:
I completely agree with you on this @gpeal. If such is the case then can you please help explain the difference between why the same is handled differently in ios. You may take the same scenario, for example client receives a url other than a valid lottie json (could be an image asset url) from backend. I've personally recently got scrutinized over the fact that why android app was crashing but not the ios app in this scenario, even though in both cases we failed to explicitly set a failure listener on both platforms. If ios component has some sort of graceful handling, why can't android component have the same? On top of that if a bad url makes its way to the component, most non-technical people commented that they'd prefer that  nothing renders on ui layer rather than the app crashing immediately. I'm open to a different approach for the solution depending upon your answer here.

gpeal:
> I completely understand this @gpeal. If such is the case then can you please help explain the difference between why the same is handled differently in ios. You may take the same scenario, for example client receives a url other than a valid lottie json (could be an image asset url) from backend. I've personally recently got scrutinized over the fact that why android app was crashing but not the ios app in this scenario, even though in both cases we failed to explicitly set a failure listener on both platforms. If ios component has some sort of graceful handling, why can't android component have the same? On top of that if a bad url makes its way to the component, most non-technical people commented that they'd prefer that nothing renders on ui layer rather than the app crashing immediately. I'm open to a different approach for the solution depending upon your answer here.

The two libraries are different because they were built and designed by different people at different times. There isn't explicitly a right and wrong choice here and iOS must have chosen differently.

You have the tools to handle failures either by using your own failure listener or by fetching the json outside of the Lottie library with your own network stack (recommended).

I'm sorry you got scrutinized but the alternative of failing silently forever usually leads to worse user experiences.

Issue #2609, opened 2025-02-11
CRASH: Fatal Exception: java.lang.IllegalStateException: Underflow in restore - more restores than saves


arcao:
Deps:
-  com.airbnb.android:lottie:6.6.2

Device Model:
- Pixel 3a XL (Android 12)

App sometime crash in animation to the destination fragment where lottie is inside layout. The fragment is newer shown and it crash during lottie rendering.  These crashes started after updating Lottie from version 6.5.2 to 6.6.2. Before there was no crashes.

```
Fatal Exception: java.lang.IllegalStateException: Underflow in restore - more restores than saves
       at android.graphics.Canvas.restore(Canvas.java:647)
       at com.airbnb.lottie.model.layer.ImageLayer.drawLayer(ImageLayer.java:100)
       at com.airbnb.lottie.model.layer.BaseLayer.draw(BaseLayer.java:270)
       at com.airbnb.lottie.model.layer.CompositionLayer.drawLayer(CompositionLayer.java:161)
       at com.airbnb.lottie.model.layer.BaseLayer.draw(BaseLayer.java:270)
       at com.airbnb.lottie.LottieDrawable.drawDirectlyToCanvas(LottieDrawable.java:1751)
       at com.airbnb.lottie.LottieDrawable.draw(LottieDrawable.java:758)
       at android.widget.ImageView.onDraw(ImageView.java:1442)
       at android.view.View.draw(View.java:22704)
       at android.view.View.updateDisplayListIfDirty(View.java:21579)
       at android.view.View.draw(View.java:22435)
       at android.view.ViewGroup.drawChild(ViewGroup.java:4528)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4289)
       at androidx.constraintlayout.widget.ConstraintLayout.dispatchDraw(ConstraintLayout.java:2082)
       at android.view.View.updateDisplayListIfDirty(View.java:21570)
       at android.view.View.draw(View.java:22435)
       at android.view.ViewGroup.drawChild(ViewGroup.java:4528)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4289)
       at android.view.View.updateDisplayListIfDirty(View.java:21570)
       at android.view.View.draw(View.java:22435)
       at android.view.ViewGroup.drawChild(ViewGroup.java:4528)
       at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4289)
       at android.view.View.draw(View.java:22707)
[...]

PR #2635, opened 2025-05-14
Fix crash when drawing shadow on image using RenderNode

allenchen1154:
When using the RenderNode rendering strategy, the call to `OffscreenLayer.finish()` decrements the save count on the `Canvas` returned from `OffscreenLayer.start()`, so we need to know whether to skip the call to `Canvas.restore()` in `ImageLayer`, otherwise the `Canvas` will thrown an exception.

Fixes https://github.com/airbnb/lottie-android/issues/2609

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
@@ -553,4 +553,12 @@ public class OffscreenLayer {
     targetCanvas.drawRenderNode(shadowRenderNode);
     targetCanvas.restore();
   }
+
+  /**
+   * Whether calling finish() will decrement the save count of the [Canvas] returned by [start()].
+   */
+  public boolean finishDecrementsCanvasSaveCount() {
+    // endRecording() will decrement the save count of the Canvas returned by beginRecording()
+    return currentStrategy == RenderStrategy.RENDER_NODE;
+  }
 }
--- a/lottie/src/main/java/com/airbnb/lottie/model/layer/ImageLayer.java
+++ b/lottie/src/main/java/com/airbnb/lottie/model/layer/ImageLayer.java
@@ -95,6 +95,9 @@ public class ImageLayer extends BaseLayer {
 
     if (renderOffScreen) {
       offscreenLayer.finish();
+      if (offscreenLayer.finishDecrementsCanvasSaveCount()) {
+        return; // Don't need a restore() call since it's handled by OffscreenLayer
+      }
     }
 
     targetCanvas.restore();


allenchen1154:
Draft for now as I would like to add a repro test case.

lottie/src/main/java/com/airbnb/lottie/utils/OffscreenLayer.java
```diff
@@ -553,4 +553,12 @@ private void renderHardwareShadow(Canvas targetCanvas, DropShadow shadow) {
     targetCanvas.drawRenderNode(shadowRenderNode);
     targetCanvas.restore();
   }
+
+  /**
+   * Whether calling finish() will decrement the save count of the [Canvas] returned by [start()].
+   */
+  public boolean finishDecrementsCanvasSaveCount() {
+    // endRecording() will decrement the save count of the Canvas returned by beginRecording()
+    return currentStrategy == RenderStrategy.RENDER_NODE;
```
allenchen1154, line 562: @geomaster I'm not sure why this happens, but I narrowed down the `beginRecording()`/`endRecording()` interaction as the root cause of the underflow error.

allenchen1154:
@gpeal How can I add a test that involves dynamically injecting an image? For reference my repro code is as follows:
```
val composition by rememberLottieComposition(
    LottieCompositionSpec
        .RawRes(R.raw.feat_campaignlisting_guestsubmission__submit_request),
)
val progress by animateLottieCompositionAsState(composition = composition, isPlaying = true)

val bitmap = Bitmap.createScaledBitmap(
    BitmapFactory.decodeResource(LocalContext.current.resources, R.drawable.feat_campaignlisting_guestsubmission__img_1),
    324,
    238,
    false,
)

val dynamicProperties = rememberLottieDynamicProperties(
    rememberLottieDynamicProperty(LottieProperty.IMAGE, bitmap, "**", "Card.png"),
)

LottieAnimation(
    composition = composition,
    progress = { progress },
    dynamicProperties = dynamicProperties,
)
```

snapshot-tests/src/androidTest/java/com/airbnb/lottie/snapshots/SnapshotTestCaseContext.kt
```diff
@@ -255,3 +256,7 @@ private suspend fun View.awaitFrame() {
         }
     }
 }
+
+internal fun SnapshotTestCaseContext.getBitmapFromAssets(name: String): Bitmap {
+    return BitmapFactory.decodeStream(context.assets.open(name), null, BitmapFactory.Options())!!
+}
```
allenchen1154, line 262: Moved this utility function so it can be shared

gpeal (approved):

PR #2640, opened 2025-06-16
Update the gradle wrapper to 8.14.2

allenchen1154:
This version supports Java versions up to 24: https://docs.gradle.org/current/userguide/compatibility.html


gpeal (approved):
Nice. I think if you run `./gradlew wrapper` a second time after upgrading, it'll upgrade the gradlew script if it has changed in a newer version of gradle.

PR #2643, opened 2025-06-20
In DefaultLottieFetchResult, catch NPE from getErrorFromConnection()

hansonchris:
The getErrorFromConnection() method attempts to construct an InputStreamReader, passing in HttpURLConnection.getErrorStream() as the constructor's arg. According to the doc for HttpURLConnection.getErrorStream(), this method can return null if there have been no errors, the connection is not connected or the server sent no useful data. The InputStreamReader(InputStream) constructor passes its arg to java.io.Reader, and that class throws an NPE if the arg is null.


gpeal:
Thanks!

hansonchris:
Definitely, thanks @gpeal for the quick merge!

PR #2660, opened 2025-09-10
Make `lottieComposition` public

thebehera:
`lottieComposition(..)` is now public.  

This is helpful when you want to ensure the LottieComposition is loaded before triggering any state changes in a suspendable function or complex LaunchedEffect


allenchen1154:
@gpeal we have use cases where we want to load a `LottieComposition` from outside of a Composable scope. Do you think exposing this function makes sense?

gpeal (approved):

thebehera:
Just updated lottie-compose.api to include the new public signatures. Hopefully that fixes the CI

gpeal (approved):

Issue #2636, opened 2025-05-14
Transparent backgrounds not working in Versions > 6.4.0


G4S-LA:
Animation have a black background color instead of transparent after upgrading the lib version from 6.4.0
Software render mode don't resolve the problem, animation just hide the other layers

**Steps to Reproduce:**

1. Use Lottie dependency version 6.4.0 – the animation works as expected.

2. Upgrade to 6.5.0 – the same animation works incorrect

**Expected Behavior:**
The Lottie animation should work the same as it did in the older version.

**Actual Behavior:**
After upgrading to 6.5.0, the animation doesn't play correctly cause have a black background color

| vesion < 6.5.0 | vesion >= 6.5.0 |  vesion >= 6.5.0 + render mode is software  |
|-------|-----|-------|
| <img width="119" alt="Image" src="https://github.com/user-attachments/assets/b58d51a2-482e-494d-abba-c1b5d440c947" /> | <img width="153" alt="Image" src="https://github.com/user-attachments/assets/a2151d4d-1626-4c13-a5b9-e1b4e52e5587" />  | <img width="144" alt="Image" src="https://github.com/user-attachments/assets/044966af-16c7-43b8-b6da-7f8c2a57c91c" /> |

Animation to reproduces this bug – [clock_animation.json](https://github.com/user-attachments/files/20204806/clock_animation.json)

rmazajev:
Can confirm, facing same issue on version 6.6.6

LudoFarkas:
I also confirm this issue. However, In my investigation, the transparent background turns to black on Android >= 14 (API level 34). On lower versions it looks OK.

LeongXianJun:
expo/react-native developer here
using lottie/react-native in my app
attempt to bump that library, which underlying bump lottie-android from 6.4.0 to 6.5.2
https://github.com/lottie-react-native/lottie-react-native/pull/1281/files

after that bump, observe similar issue

collin-mumey:
I've also experienced this issue. It appears to be related to the multiply changes that were introduced in version 6.5.0. After removing all multiply effects from our animation it renders properly on the device (API level 36).

allenchen1154:
This is fixed in [version 6.6.9](https://github.com/airbnb/lottie-android/releases/tag/v6.6.9)

Issue #2593, opened 2024-12-11
Rendering (clipping) issue in 6.5.0+


kaidotarmahcl:
**Describe the bug**
Part of the animation is clipped since library version 6.5.0+.
The issue might be related to https://github.com/airbnb/lottie-android/issues/2587

**What version of Lottie did you test this on?**
All versions from 6.5.0+ had the same issue. 
6.1.0 - 6.4.0 works correctly. Did not test earlier versions.

**What version of Android did you test this on?**
Android 12, Android 15

**Steps To Reproduce**
In library version 6.4.0 the animation works correctly.
In library version 6.5.0+ the animation does not work correctly and some clipping happens.

**Recordings**
6.4.0 - working correctly
https://github.com/user-attachments/assets/bc736f33-53af-4bc6-be32-138967d990af

6.5.0 - working incorrectly
https://github.com/user-attachments/assets/b4354d79-0616-4f6f-94de-f36a05e3fb2a

gpeal:
Can you attach an animation that reproduces this and can you try setting the render mode to software?

kaidotarmahcl:
@gpeal Here is the problematic file:
[ill_onb_processing_full.lottie.zip](https://github.com/user-attachments/files/18146620/ill_onb_processing_full.lottie.zip)

Enabling software rendering did not change anything.

matthieu-rollin:
Confirming that the last working version was `6.4.1`, any other upper releases get clipped rendering like the example in this thread.

kaidotarmahcl:
@gpeal Hi, just wanted to check in. Anything new on this? Let me know if there's anything I can do to help.

smetz:
For an extra data point, I'm also seeing what seems to be this clipping issue with [this animation](https://github.com/user-attachments/files/18455815/coin_day_30.json)

I've tested multiple Android versions and am only seeing the clipping problem on API 34 and above (tried 30, 33, 34 and 35)
I have also tested Lottie versions 6.1.0-6.6.2 and like the previous report, the clipping seems to only happen on 6.5.0+
Changing the `renderMode` setting to `RenderMode.SOFTWARE` also didn't fix it for me

cyprian:
In React Native ("lottie-react-native": "7.2.1") setting renderMode="SOFTWARE" fixed the clipping for me. 

```
<LottieView style={$animation} 
        source={require("../../assets/animations/logo-animation.json")}
        resizeMode="contain"
[...]

PR #2661, opened 2025-09-13
Fix transparent background rendering with MULTIPLY Blend mode on Android 34+

allenchen1154:
The workaround in https://github.com/airbnb/lottie-android/pull/2519 for Multiply blend modes doesn't work on Android 34+. This changes the implementation to use `BlemdModeCompat.MULTIPLY` on Android versions Q (29) and higher where it is supported, while retaining the workaround for older versions.

Fixes https://github.com/airbnb/lottie-android/issues/2636 and https://github.com/airbnb/lottie-android/issues/2593

|Before|After|
|-|-|
|Screenshot_20250912_214839|Screenshot_20250912_214546|

|Before|After|
|-|-|
|Screenshot_20250912_214850|Screenshot_20250912_214600|


allenchen1154:
Note: This bug is only reproducible when using Hardware rendering.

allenchen1154:
cc @geomaster

PR #2662, opened 2025-09-24
Safeguard against very large Bitmap creation

allenchen1154:
Because we rely on the transform `Matrix` returned from the `Canvas` to size a `Bitmap` when using software rendering, an erroneous matrix value can cause a very large Bitmap to be allocated, triggering an OutOfMemory exception.

We've encountered such a situation when using `layoutlib` for screenshot tests, which uses a `NopCanvas` [on initial render](https://cs.android.com/android/_/android/platform/frameworks/layoutlib/+/7b05b277beee599532606e9bb6d7a71f5ca2ab6e:bridge/src/com/android/layoutlib/bridge/impl/RenderSessionImpl.java;l=543;bpv=1;bpt=0;drc=085857f145aac790e2a08cf6eb9546f98e26c338) that can return an invalid `Matrix.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
@@ -111,6 +111,11 @@ public class LottieDrawable extends Drawable implements Drawable.Callback, Anima
       "reducedmotion"
   );
 
+  /**
+   * Significantly larger than the largest expected size, equivalent to about 1.5x an 8K display.
+   */
+  private static final long MAX_SOFTWARE_BITMAP_PIXELS = 50_000_000L;
+
   private LottieComposition composition;
   private final LottieValueAnimator animator = new LottieValueAnimator();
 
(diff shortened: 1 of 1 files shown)


lottie/src/main/java/com/airbnb/lottie/LottieDrawable.java
```diff
@@ -1798,7 +1798,8 @@ private void renderAndDrawAsBitmap(Canvas originalCanvas, CompositionLayer compo
     int renderWidth = (int) Math.ceil(softwareRenderingTransformedBounds.width());
     int renderHeight = (int) Math.ceil(softwareRenderingTransformedBounds.height());
 
-    if (renderWidth <= 0 || renderHeight <= 0) {
+    // Safeguard against errors during Bitmap creation by returning early if dimensions are invalid.
+    if (renderWidth <= 0 || renderHeight <= 0 || renderWidth > bounds.width() || renderHeight > bounds.height()) {
```
gpeal, line 1802: Wouldn't this not render an image if it scaled up so it covered the whole screen + 1px?

allenchen1154: In that case can we clip the bounds so that we only allocate a bitmap large enough to draw what's shown onscreen?
I'm trying to guard against cases where the Canvas matrix returns an invalid scale or translation that causes something like a 1,000,000 x 1,000,000 bitmap to be allocated.

gpeal:
Could you take a look at the snapshot tests? Looks like there are a bunch of blanks now.

PR #2664, opened 2025-10-07
Add 3D rotation support for Lottie animations

hepiao3:
The purpose is to solve the problem of Android Lottie not supporting 3D layers.


eval3:
todo fix

PR #2674, opened 2025-10-27
v6.7.0-snapshot

eval3:


eval3:
@gpeal Can this automatically trigger the publishing process？

PR #2706, opened 2026-03-17
Lottie: Validate Looper compatibility when retrieving from taskCache

fengjial-google:
This is a follow up of https://github.com/airbnb/lottie-android/pull/2685

The static taskCache in LottieCompositionFactory can store LottieTasks created on different UI threads (e.g., the system main thread vs. a per-window UI thread like the Taskbar thread).

Previously, if an animation was requested on the Taskbar UI thread that had already been cached by the Main thread, the Factory would return the cached task. This task would then notify listeners on the wrong thread, leading to a CalledFromWrongThreadException when LottieAnimationView attempted to requestLayout().

This change:
1. Adds getLooper() to LottieTask to expose its target notification thread.
2. Updates LottieCompositionFactory.cache() to only reuse a cached task if its Looper matches the requested uiLooper.

### DIFF
--- a/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
+++ b/lottie/src/main/java/com/airbnb/lottie/LottieCompositionFactory.java
@@ -816,7 +816,11 @@ public class LottieCompositionFactory {
       task = new LottieTask<>(uiLooper, cachedComposition);
     }
     if (cacheKey != null && taskCache.containsKey(cacheKey)) {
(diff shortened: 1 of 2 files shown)


fengjial-google:
@gpeal can I get a review as this is related to https://github.com/airbnb/lottie-android/pull/2685

fengjial-google:
bump @gpeal

fengjial-google:
bump again @gpeal

fengjial-google:
@gpeal bump again

PR #2720, opened 2026-05-26
Preserve text color alpha

RickAi:
## Summary
- Preserve the alpha component from text fill and stroke colors when applying layer, text range, and parent alpha.
- Prevent text layers with missing fill or stroke colors from becoming visible black after alpha is applied.
- Add coverage for transparent missing colors, explicit opaque colors, and combined alpha.

## Reproduction
A Lottie document with a text layer that omits `fc` but includes `sc` and `sw` should render without a black fill. Before this fix, the missing fill color is initialized as transparent black and becomes visible black when layer alpha is applied.

## Screenshots
Attached in the PR discussion: AE expected rendering and Lottie Android rendering before the fix.

## Test
- `./gradlew :lottie:testDebugUnitTest --tests com.airbnb.lottie.model.layer.TextLayerTest`


RickAi:
Screenshots for the repro:

| AE expected rendering | Lottie Android before this fix |
| --- | --- |
| AE expected rendering | Lottie Android before this fix |

RickAi:
Reproduction Lottie JSON:

[data.json](https://github.com/user-attachments/files/28265367/data.json)

anandfresh (approved):

PR #2723, opened 2026-06-11
Add lightweight system tracing option

nicomazz:
This PR introduces a new lightweight tracing option (enableLightTracing) for Lottie, allowing developers to monitor all Lottie animators receiving per-frame callbacks in Perfetto traces with minimal performance overhead.

Context: In Android SystemUI, we recently encountered several Lottie-related performance issues where animations were unintentionally running indefinitely due to caller bugs. While Lottie already supports tracing, enabling the full suite of Lottie trace markers proved to be too heavy for continuous monitoring on lab and field devices.

By introducing this lightweight tracing alternative, we were successfully able to spot and fix multiple animation leaks in production without the severe overhead of the full tracing suite.

(Upstreamed from AOSP)


nicomazz:
@gpeal , could you help reviewing this? right now we have this commit in the aosp mirror of Lottie. If we merge it we avoid diverging

